Issue #16, opened 2023-01-03
Ability to test whether warnings are raised during test steps


bendbennett:
This issue is a duplicate of [Ability to test whether warnings are raised during test steps](https://github.com/hashicorp/terraform-plugin-testing/issues/69). 

### Version
N/A

### Use-cases
Currently, ExpectError [field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/resource#TestStep) only allows checking errors, not warnings. This makes provider developers hard to test certain scenarios where warnings are expected and want to validate the exact behavior in tests.

### Attempted Solutions
N/A

### Proposal
Add an `ExpectWarning` field to `TestStep` and use json output from terraform commands to determine whether warnings have been generated during execution of a test step.

Requires changes to [terraform-exec](https://github.com/hashicorp/terraform-exec) to add support for executing terraform commands with `-json` option (see References below for links to relevant issue and PRs on [terraform-exec repository](https://github.com/hashicorp/terraform-exec).

### References
- https://discuss.hashicorp.com/t/how-to-test-warnings/34259
- [Plan/Apply -json support missing](https://github.com/hashicorp/terraform-exec/issues/183#top)
- [Add JSON option for "plan" and "apply"](https://github.com/hashicorp/terraform-exec/pull/275#top)
- [json log for plan and apply](https://github.com/hashicorp/terraform-exec/pull/329#top)

bflad:
Closing as a duplicate of the (now transferred) #69.

PR #17, opened 2023-01-03
Add ability to test whether warnings are raised during test steps

bendbennett:
Closes: #16


helper/resource/testing_new.go
```diff
@@ -120,6 +120,9 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
 	// acts as default for import tests
 	var appliedCfg string
 
+	// stdout (io.Writer) is supplied to all terraform commands that are using streaming json output.
+	stdout := NewStdout()
```
bflad, line 124: I think we should have each command use its own buffer for a few reasons:

- Prevents the need for everywhere to pass it and for the business logic to deal with any buffering aspects
- Prevents any strange cross-command contamination
- Allows the buffer to be garbage collected, rather than potentially keeping a buffer around that's the largest size from any of the commands it was used in

bendbennett: Have made a first pass at this. Interested to hear your thoughts.

I've also left a couple of "TODO"s. Would value your input on these.

internal/plugintest/terraform_json_buffer.go
```diff
@@ -0,0 +1,125 @@
...
+var _ io.Reader = &TerraformJSONBuffer{}
+
+// TerraformJSONBuffer is used for storing and processing streaming
+// JSON output generated when running terraform commands with the
+// `-json` flag.
+type TerraformJSONBuffer struct {
+	buf         *bytes.Buffer
+	diagnostics TerraformJSONDiagnostics
+	jsonOutput  []string
+	rawOutput   string
```
bendbennett, line 42: `rawOutput` was added in order to capture any outputs that were not JSON in order to maximise the information being returned to the provider developer for the purposes of debugging.

For example:
```shell

{"@level":"info","@message":"Terraform 1.3.2","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.322845Z","terraform":"1.3.2","type":"version","ui":"1.0"}
{"@level":"warn","@message":"Warning: Empty or non-existent state","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.355523Z","diagnostic":{"severity":"warning","summary":"Empty or non-existent state","detail":"There are currently no remote objects tracked in the state, so there is nothing to refresh."},"type":"diagnostic"}
{"@level":"info","@message":"Outputs: 0","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.355853Z","outputs":{},"type":"outputs"}
{"@level":"info","@message":"Terraform 1.3.2","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.380962Z","terraform":"1.3.2","type":"version","ui":"1.0"}
{"@level":"warn","@message":"Warning: Provider development overrides are in effect","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.381655Z","diagnostic":{"severity":"warning","summary":"Provider development overrides are in effect","detail":"The following provider development overrides are set in the CLI configuration:\n - bendbennett/playground in /Users/bdb/go/bin\n - bendbennett/timeouts in /Users/bdb/go/bin\n\nThe behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases."},"type":"diagnostic"}
{"@level":"info","@message":"example_resource.test: Plan to create","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.396541Z","change":{"resource":{"addr":"example_resource.test","module":"","resource":"example_resource.test","implied_provider":"example","resource_type":"example_resource","resource_name":"test","resource_key":null},"action":"create"},"type":"planned_change"}
{"@level":"info","@message":"Plan: 1 to add, 0 to change, 0 to destroy.","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.396576Z","changes":{"add":1,"change":0,"remove":0,"operation":"plan"},"type":"change_summary"}

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # example_resource.test will be created
  + resource "example_resource" "test" {
      + id = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.
{"@level":"info","@message":"Terraform 1.3.2","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.479418Z","terraform":"1.3.2","type":"version","ui":"1.0"}
{"@level":"warn","@message":"Warning: Provider development overrides are in effect","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.480270Z","diagnostic":{"severity":"warning","summary":"Provider development overrides are in effect","detail":"The following provider development overrides are set in the CLI configuration:\n - bendbennett/playground in /Users/bdb/go/bin\n - bendbennett/timeouts in /Users/bdb/go/bin\n\nThe behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases."},"type":"diagnostic"}
{"@level":"info","@message":"example_resource.test: Plan to create","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.485951Z","change":{"resource":{"addr":"example_resource.test","module":"","resource":"example_resource.test","implied_provider":"example","resource_type":"example_resource","resource_name":"test","resource_key":null},"action":"create"},"type":"planned_change"}
{"@level":"info","@message":"example_resource.test: Creating...","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.491701Z","hook":{"resource":{"addr":"example_resource.test","module":"","resource":"example_resource.test","implied_provider":"example","resource_type":"example_resource","resource_name":"test","resource_key":null},"action":"create"},"type":"apply_start"}
{"@level":"info","@message":"example_resource.test: Creation errored after 0s","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.492349Z","hook":{"resource":{"addr":"example_resource.test","module":"","resource":"example_resource.test","implied_provider":"example","resource_type":"example_resource","resource_name":"test","resource_key":null},"action":"create","elapsed_seconds":0},"type":"apply_errored"}
{"@level":"error","@message":"Error: error summary","@module":"terraform.ui","@timestamp":"2023-01-25T13:54:15.503665Z","diagnostic":{"severity":"error","summary":"error summary","detail":"error detail","address":"example_resource.test","range":{"filename":"terraform_plugin_test.tf","start":{"line":1,"column":36,"byte":35},"end":{"line":1,"column":37,"byte":36}},"snippet":{"context":"resource \"example_resource\" \"test\"","code":"resource \"example_resource\" \"test\" { }","start_line":1,"highlight_start_offset":35,"highlight_end_offset":36,"values":[]}},"type":"diagnostic"}
```

internal/plugintest/terraform_json_buffer.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return b.buf.Read(p)
+}
+
+func (b *TerraformJSONBuffer) Parse() {
+	if b.buf == nil {
+		log.Fatal("call NewTerraformJSONBuffer to initialise buffer")
+	}
+
+	scanner := bufio.NewScanner(b.buf)
```
bflad, line 73: Do we need to worry about the 64KB [`bufio.MaxScanTokenSize`](https://pkg.go.dev/bufio#MaxScanTokenSize)? If Terraform can send lines larger than the default buffer size, we may need to provide our own to match Terraform's largest possible line size.

internal/plugintest/terraform_json_buffer_test.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			t.Errorf("cannot write to tfJSON: %s", err)
+		}
+
+		fileEntries = append(fileEntries, txt)
+	}
+
+	if err := scanner.Err(); err != nil {
+		t.Errorf("scanner error: %s", err)
+	}
```
bflad, line 114: Since `os.File` is itself an `io.Reader`, can it just be copied to `tfJSON`? e.g.

```go
	tfJSON := NewTerraformJSONBuffer()

	file, err := os.Open("../testdata/terraform-json-output.txt")
	if err != nil {
		t.Errorf("cannot read file: %s", err)
	}
	defer file.Close()

	_, err := io.Copy(tfJSON, file)

	if err != nil {
		t.Fatalf("cannot copy file contents to buffer: %s", err)
	}
```

I think then we could take that a step further if we wanted so you could potentially create the `TerraformJSONBuffer` using file contents for when we get more test cases:

```go
func NewTerraformJSONBufferFromFile(path string) error {
	tfJSON := NewTerraformJSONBuffer()

	file, err := os.Open(path)

	if err != nil {
		return fmt.Errorf("cannot read file: %w", err)
	}

	defer file.Close()

	_, err := io.Copy(tfJSON, file)

	if err != nil {
		return fmt.Errorf("cannot copy file contents to buffer: %w", err)
	}

	return nil
}
```

Or conversely it could just be a method such as `ReadFile()`:

```go
func (b TerraformJSONBuffer) ReadFile(path string) error {
	file, err := os.Open(path)

	if err != nil {
		return fmt.Errorf("cannot read file: %w", err)
	}

	defer file.Close()

	_, err := io.Copy(b, file)

	if err != nil {
		return fmt.Errorf("cannot copy file contents to buffer: %w", err)
	}

	return nil
}
```

internal/plugintest/terraform_json_buffer.go
bflad: Could you expand a little more on when this output might be shown to provider developers? At first glance I'd be worried about showing this level of detail unless its potentially in TRACE logging from the testing code. There are separate environment variables for outputting Terraform core's logging, although they can be problematic sometimes.

I don't think it's necessarily a bad idea to capture this data though in case we do need it for advanced troubleshooting cases.

internal/plugintest/working_dir.go
```diff
@@ -169,15 +171,64 @@ func (wd *WorkingDir) planFilename() string {
...
+		return createPlanResponse, err
+	}
+
+	if !hasChanges {
+		logging.HelperResourceTrace(ctx, "Created plan with no changes")
+
+		return createPlanResponse, nil
+	}
+
+	stdout, err := wd.SavedPlanRawStdout(ctx)
```
bflad, line 212: Should this `stdout` be wired with the `CreatePlanResponse` stdout? Maybe the overlappting naming is confusing me. 😄

bflad (commented):
Maybe it's easiest to chat through this one early next week? I think its getting closer, but I'm getting a little lost in the details on a Friday. 👍

internal/plugintest/terraform_json_buffer.go
bendbennett: Have switched to using [bufio.NewReader](https://pkg.go.dev/bufio#NewReader).

internal/plugintest/terraform_json_buffer_test.go
bendbennett: Have incorporated both `NewTerraformJSONBufferFromFile` and `ReadFile`. Thanks!

internal/plugintest/working_dir.go
bendbennett: I have removed the usage of stdout from `CreatePlanResponse` and the other `...Response`(s) as it does not seem to be necessary or desirable to show this level of information to provider developers.

helper/resource/testing_new.go
```diff
@@ -253,34 +254,107 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
...
 			} else {
 				if err != nil && c.ErrorCheck != nil {
 					logging.HelperResourceDebug(ctx, "Calling TestCase ErrorCheck")
+
+					// This is a "best effort" to supply ErrorCheck with the contents of the
+					// error diagnostics returned from running the Terraform command when
+					// executed with the -json flag. If ErrorCheck functions are using
+					// regexp to perform matches that include line breaks then these will
+					// likely no longer behave as expected and therefore potentially
+					// represent a breaking change.
```
bflad, line 300: For others benefit, the `TestCase.ErrorCheck` functionality is the main wrinkle that prevents us from immediately being able to use Terraform's machine readable output (JSON view/UI).

`ErrorCheck` functionality, currently used in many popular Terraform provider acceptance tests, was intended to serve these use cases:

- Skipping tests based on known errors (e.g. an API error indicating that a feature is not supported against an endpoint variant)
- Rewriting errors for clarity
- Ignoring known errors

Today the testing code is reliant on terraform-exec's `error` returns to include the data needed for `TestCase.ErrorCheck` and `TestStep.ExpectError` functionality to work. Warning diagnostics are not passed back the same way, since they are not errors and should not stop program execution. In order to capture similar data for warnings, the testing code would either need to perform error-prone (and likely version dependent) scraping of the human-readable output or switch to the machine-readable output. The machine-readable output means that the `error` from terraform-exec needs to be fundamentally treated differently and trying to convert machine-readable error diagnostics to be functionally equivalent to the prior `error` would be a non-trivial exercise that's not guaranteed to fully keep backwards compatibility.

An approach to this situation may be to prevent defining both `TestCase.ErrorCheck` and the new `TestStep.ExpectWarning` in the same `TestCase`, and setting up the testing execution to continue using the human-readable output where `ErrorCheck` is used. This is a non-starter for providers such as the Terraform AWS Provider though, which defines `ErrorCheck` functionality on _all_ tests, meaning `ExpectWarning` could never be used.

Definitely open to ideas here, but trying to work around this further is risky at best.

Issue #18, opened 2023-01-03
Option to Retain Acceptance Testing Files


bendbennett:
This issue is a duplicate of [Option to Retain Acceptance Testing Files](https://github.com/hashicorp/terraform-plugin-sdk/issues/701#top).

### SDK version
N/A

### Use-cases

_This issue is created by request as part of a series of related acceptance test feature requests that will be cross-referenced to each other for discoverability._

Currently the acceptance testing framework always deletes its temporary directory full of all the intermediate Terraform CLI goodies that happened during a test. The ability to temporarily save these files may aid in the troubleshooting of the acceptance testing framework itself, viewing of finalized test configurations, and potentially debugging of provider issues (such as inspecting intermediate plans or states).

### Attempted Solutions

None at the moment.

### Proposal

Add an environment variable to skip the deletion of the temporary directory created by the acceptance testing framework. The debug log or `t.Log()` equivalent could output the root path to view those files.

### References

- https://github.com/hashicorp/terraform-plugin-sdk/issues/698
- https://github.com/hashicorp/terraform-plugin-sdk/issues/699
- https://github.com/hashicorp/terraform-plugin-sdk/issues/700

PR #19, opened 2023-01-03
Add Option to Retain Acceptance Testing Files

bendbennett:
Closes: #18


helper/resource/testing_new.go
```diff
@@ -442,3 +452,27 @@ func testIDRefresh(ctx context.Context, t testing.T, c TestCase, wd *plugintest.
 
 	return nil
 }
+
+func copyWorkingDir(ctx context.Context, t testing.T, stepNumber int, wd *plugintest.WorkingDir) {
+	if v := os.Getenv(plugintest.EnvTfAccPersistWorkingDir); v != "" {
+		dest := wd.GetHelper().WorkingDirectory() + "_" + strconv.Itoa(stepNumber)
+
+		err := os.Setenv(plugintest.EnvTfAccPersistWorkingDirPath, wd.GetHelper().WorkingDirectory())
```
bflad, line 460: Is this new environment variable just for testing the functionality? I'm wondering if there is a different way we can achieve this. Maybe not for this pull request, but currently we randomly generate where the working directory is located. Instead, maybe the working directory could be passed into the test case (falling back to the random directory), so consumers (and our testing 😄 ) can choose where it goes. e.g.

```go
func TestXXX(t *testing.T) {
  t.Setenv(plugintest.EnvTfAccPersistWorkingDir, "1")
  workDir := t.TempDir()

  resource.TestCase{
    // WorkingDir sets the base directory where testing files used by the testing
    // module. It defaults to a randomized, temporary directory.
    WorkingDir: workDir,
  }

  // Checks against workDir
}
```

helper/resource/teststep_providers_test.go
```diff
@@ -1871,6 +1874,70 @@ func TestTest_TestStep_ProviderFactories_Refresh_Inline(t *testing.T) {
 	})
 }
 
+func TestTest_TestStep_ProviderFactories_CopyWorkingDir_EachTestStep(t *testing.T) {
+	t.Parallel()
+
+	err := os.Setenv(plugintest.EnvTfAccPersistWorkingDir, "1")
+	if err != nil {
+		t.Fatalf("cannot set %s env var: %s", plugintest.EnvTfAccPersistWorkingDir, err)
+	}
```
bflad, line 1883: Similar to `t.TempDir()`, there's [`(*testing.T).Setenv()`](https://pkg.go.dev/testing#T.Setenv). Setting environment variables in tests should implicitly _not_ be parallelized since any other places that get/set the environment variable would result in race conditions.

bflad (commented):
Shaping up nicely -- please reach out with questions.

helper/resource/copy.go
```diff
@@ -0,0 +1,69 @@
+package resource
+
+import (
+	"io"
+	"os"
+	"path"
+)
+
+// CopyFile copies a single file from src to dest.
+func CopyFile(src, dest string) error {
```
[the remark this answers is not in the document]
bendbennett: Have moved functions that were in `helper/resource/copy.go` into `internal/plugintest/util.go`.

helper/resource/testing.go
```diff
@@ -418,6 +418,11 @@ type TestCase struct {
 	// IDRefreshIgnore is a list of configuration keys that will be ignored
 	// during ID-only refresh testing.
 	IDRefreshIgnore []string
+
+	// WorkingDir sets the base directory where testing files used by the testing
+	// module are generated. If WorkingDir is unset, a randomized, temporary
+	// directory is used.
```
bflad, line 424: This might be a good place to also call out the new environment variable 👍 

```suggestion
	// WorkingDir sets the base directory where testing files used by the testing
	// module are generated. If WorkingDir is unset, a randomized, temporary
	// directory is used.
	//
	// Use the TF_ACC_PERSIST_WORKING_DIR environment variable, conventionally
	// set to "1", to persist any working directory files. Otherwise, this directory is
	// automatically cleaned up at the end of the TestCase.
```

helper/resource/testing.go
```diff
@@ -418,6 +418,11 @@ type TestCase struct {
...
+	WorkingDir string
```
bflad, line 425: This awesome new field should get a new CHANGELOG entry as well 😄

helper/resource/testing_new.go
```diff
@@ -442,3 +453,25 @@ func testIDRefresh(ctx context.Context, t testing.T, c TestCase, wd *plugintest.
...
+	err := plugintest.CopyDir(wd.GetHelper().WorkingDirectory(), dest)
+	if err != nil {
+		logging.HelperResourceError(ctx,
+			"Unexpected error copying working directory files",
+			map[string]interface{}{logging.KeyError: err},
+		)
+		t.Fatalf("TestStep %d/%d error copying working directory files: %s", stepNumber, err)
+	}
+
+	t.Log(fmt.Sprintf("Working directory and files have been copied to: %s", dest))
```
bflad, line 476: Nit: https://pkg.go.dev/testing#T.Logf (surprised that this isn't caught by `go vet` -- maybe we should double check if there's an upstream issue for that?)

```suggestion
	t.Logf("Working directory and files have been copied to: %s", dest)
```

helper/resource/teststep_providers_test.go
```diff
@@ -1778,6 +1782,54 @@ func TestTest_TestStep_ProviderFactories_Import_External_WithPersistMatch(t *tes
...
+		WorkingDir: workingDir,
+		Steps:      testSteps,
+	})
+
+	workingDirPath := filepath.Dir(workingDir)
+
+	for k := range testSteps {
+		dir := workingDirPath + "_" + strconv.Itoa(k+1)
+
+		_, err := os.ReadDir(dir)
```
bflad, line 1826: Open question: Should we try to verify any of the contents within the directories? I think for at least one of the tests it might be good to verify that we get:

- Step 1: configuration file and state file
- Step 2: configuration file, plan file, and state file

internal/plugintest/util.go
```diff
@@ -68,3 +70,65 @@ func symlinkDirectoriesOnly(srcDir string, destDir string) error {
...
 	return nil
 }
+
+// CopyFile copies a single file from src to dest.
+func CopyFile(src, dest string) error {
+	var srcFileInfo os.FileInfo
+
+	srcFile, err := os.Open(src)
+	if err != nil {
+		return err
```
bflad, line 80: Nit: It might be good to wrap these errors (here and below) with what we were trying to do so folks can troubleshoot any issues easier:

```go
		return fmt.Errorf("unable to open file: %w", err)
```

website/docs/plugin/testing/acceptance-tests/index.mdx
```diff
@@ -236,37 +236,38 @@ jobs:
...
+| `TF_ACC_TERRAFORM_VERSION`   | N/A                                                                           | Set the exact version of Terraform CLI to automatically install into `TF_ACC_TEMP_DIR`. For example, `1.1.6` or `v1.0.11`.                                                                                                                                                                                                                                                                                                                                           |
+| `TF_ACC_PERSIST_WORKING_DIR` | N/A                                                                           | Set to any value to enable persisting the working directory and the files generated during execution of each `TestStep`. The location of each directory is output in the log for each `TestStep`.                                                                                                                                                                                                                                                                    |
```
bflad, line 248: We could also consider setting up actual DEBUG/INFO logging entries as well within `copyWorkingDir` -- I do actually like the current `t.Logf()` usage though so folks don't _have_ to enable `TF_LOG` or friends.

```suggestion
| `TF_ACC_PERSIST_WORKING_DIR` | N/A                                                                           | Set to any value to enable persisting the working directory and the files generated during execution of each `TestStep`. The location of each directory is written to the test output for each `TestStep` when the `go test -v` (verbose) flag is provided.                                                                                                                                                                                                                                                                    |
```

bflad (commented):
I think after this next round this will be good to go! 🎉 Excellent work.

helper/resource/teststep_providers_test.go
bendbennett: I think this is a good shout. I've added some checks to verify that config, state and plan files exist where appropriate.

helper/resource/teststep_providers_test.go
```diff
@@ -1778,6 +1782,82 @@ func TestTest_TestStep_ProviderFactories_Import_External_WithPersistMatch(t *tes
...
+			"terraform.tfstate",
+			"tfplan",
+		}
+
+		for i, file := range configPlanStateFiles {
+			// Skip verifying plan for first test step as there is no plan file if the
+			// resource does not already exist.
+			if k == 0 && i > 1 {
+				break
+			}
```
bflad, line 1852: Nit: These may be a little easier to read if the check was (substituting `k` for `testStepIndex`)

```go
if testStepIndex == 0 && file == "tfplan" {
  continue
}
```

helper/resource/teststep_providers_test.go
```diff
@@ -1811,6 +1891,83 @@ func TestTest_TestStep_ProviderFactories_Import_External_WithoutPersistNonMatch(
...
+			"tfplan",
+		}
+
+		for i, file := range configPlanStateFiles {
+			// Skip verifying state and plan for first test step as ImportStatePersist is
+			// false so the state is not persisted and there is no plan file if the
+			// resource does not already exist.
+			if k == 0 && i > 0 {
+				break
+			}
```
bflad, line 1962: Nit: Similarly

```go
if testStepIndex == 0 && (file == "terraform_plugin_test.tf" || file == "terraform.tfstate") {
  continue
}
```

bflad (approved):
Looks good to me 🚀

PR #31, opened 2023-01-18
Create website page for migrating acceptance testing

SBGoods:


website/docs/plugin/testing/migrating.mdx
```diff
@@ -16,8 +16,18 @@ This guide provides information and examples for most common use cases, but it d
...
 
 - Change all instances of the following Go import statements in `*_test.go` files:
-    - github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest to github.com/hashicorp/terraform-plugin-testing/helper/acctest
-    - github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource to github.com/hashicorp/terraform-plugin-testing/helper/resource
-- Run `go get github.com/hashicorp/terraform-plugin-testing@v1.0.0`
-- Run `go mod tidy`
-- Verify everything is working as expected.
+    - `github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest` to `github.com/hashicorp/terraform-plugin-testing/helper/acctest`
+    - `github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource` to `github.com/hashicorp/terraform-plugin-testing/helper/resource`
+- Get and download terraform-plugin-testing `v1.0.0`:
```
bendbennett, line 21: Perhaps suggest using `latest` rather than specific version?

website/docs/plugin/testing/migrating.mdx
```diff
@@ -16,8 +16,18 @@ This guide provides information and examples for most common use cases, but it d
...
+
+```shell
+$ go get github.com/hashicorp/terraform-plugin-testing@v1.0.0
```
bendbennett, line 24: Perhaps suggest using `latest` rather than specific version?

bendbennett (approved):
Looks good 👍  
Just a minor comment.

website/docs/plugin/testing/migrating.mdx
SBGoods: That's a good point, although pointing to `v1.0.0` ensures that there would be no breaking changes in the migration. Maybe we can add an additional optional step for those who want to upgrade to the latest version after making sure that the migration to `v1.0.0` works as expected?

website/docs/plugin/testing/migrating.mdx
```diff
@@ -0,0 +1,33 @@
...
+This guide provides information and examples for most common use cases, but it does not discuss every nuance of migration. You can ask additional migration questions in the [HashiCorp Discuss forum](https://discuss.hashicorp.com/c/terraform-providers/tf-plugin-sdk/43). To request additions or updates to this guide, submit issues or pull requests to the [`terraform-plugin-testing` repository](https://github.com/hashicorp/terraform-plugin-testing).
+
+
+## Migration steps
+
+Take the following steps when you migrate a provider's acceptance tests from SDKv2 to the testing module:
+
+- Change all instances of the following Go import statements in `*_test.go` files:
+    - `github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest` to `github.com/hashicorp/terraform-plugin-testing/helper/acctest`
+    - `github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource` to `github.com/hashicorp/terraform-plugin-testing/helper/resource`
```
bflad, line 20: Some folks may be tangentially pulling in the `terraform` package as well (e.g. for custom `TestCheckFunc`), so it'll be good to give that a shout as well:

```suggestion
    - `github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource` to `github.com/hashicorp/terraform-plugin-testing/helper/resource`
    - `github.com/hashicorp/terraform-plugin-sdk/v2/terraform` to `github.com/hashicorp/terraform-plugin-testing/terraform`
```

bflad (approved):
Looks good to me 🚀  I know this repository is not hooked up to the website workflows yet, but we'll want to be sure it gets added to the navigation file when that's created since this page is not referenced elsewhere.

website/docs/plugin/testing/migrating.mdx
bflad: Theoretically `github.com/hashicorp/terraform-plugin-testing@latest` should always be safe because:

- Go's import system uses separate import paths for new major versions, so v2 would be `github.com/hashicorp/terraform-plugin-testing/v2@...`
- All versions of v1 should have compatibility promises for existing functionality

PR #34, opened 2023-01-24
Create nav data file for website

SBGoods:
Adds `plugin-testing-nav-data.json` to enable lefthand navigation for the Testing page on the developer website.

PR #35, opened 2023-01-25
Add necessary website files

nayres:
This branch adds necessary website files to migrate content from terraform-testing-sdk to terraform-plugin-testing.


website/Makefile
```diff
@@ -0,0 +1,60 @@
+######################################################
+# NOTE: This file is managed by the Digital Team's   #
+# Terraform configuration @ hashicorp/mktg-terraform #
+######################################################
+
+.DEFAULT_GOAL := website
+
+# Set the preview mode for the website shell to "developer" or "io"
+PREVIEW_MODE ?= developer
+REPO ?= terraform-plugin-sdk
```
BRKalow, line 10:
```suggestion
REPO ?= terraform-plugin-testing
```

website/Makefile
```diff
@@ -0,0 +1,60 @@
...
+PWD=$$(pwd)
+
+DOCKER_IMAGE="hashicorp/dev-portal"
+DOCKER_IMAGE_LOCAL="dev-portal-local"
+DOCKER_RUN_FLAGS=-it \
+		--publish "3000:3000" \
+		--rm \
+		--tty \
+		--volume "$(PWD)/docs:/app/docs" \
+		--volume "$(PWD)/img:/app/public" \
```
BRKalow, line 26:
```suggestion
		--volume "$(PWD)/img:/app/public/img" \
```

website/data/plugin-testing-nav-data.json
```diff
@@ -0,0 +1,150 @@
+[
```
BRKalow, line 1: I think we can either exclude this file or leave it empty for now 👍

website/scripts/website-build.sh
```diff
@@ -0,0 +1,52 @@
...
+######################################################
+
+# Repo which we are cloning and executing npm run build:deploy-preview within
+REPO_TO_CLONE=dev-portal
+# Set the subdirectory name for the base project
+PREVIEW_DIR=website-preview
+# The directory we want to clone the project into
+CLONE_DIR=website-preview
+# The product for which we are building the deploy preview
+PRODUCT=terraform-plugin-sdk
```
BRKalow, line 13:
```suggestion
PRODUCT=terraform-plugin-testing
```

BRKalow (commented):
@thiskevinwang Do we also need to adjust the root `Makefile`?

website/data/plugin-testing-nav-data.json
bflad: Refer also to https://github.com/hashicorp/terraform-plugin-testing/pull/34

thiskevinwang:
> @thiskevinwang Do we also need to adjust the root `Makefile`?

Yea, the root `Makefile` should have targets that call out to `website/Makefile`

Example: https://github.com/hashicorp/terraform-plugin-framework/blob/main/Makefile

nayres:
> > @thiskevinwang Do we also need to adjust the root `Makefile`?
> 
> Yea, the root `Makefile` should have targets that call out to `website/Makefile`
> 
> Example: https://github.com/hashicorp/terraform-plugin-framework/blob/main/Makefile



> > @thiskevinwang Do we also need to adjust the root `Makefile`?
> 
> Yea, the root `Makefile` should have targets that call out to `website/Makefile`
> 
> Example: https://github.com/hashicorp/terraform-plugin-framework/blob/main/Makefile

I can make this update. Would I have to point to the website/Makefile in a specific way?

thiskevinwang:
> I can make this update. Would I have to point to the website/Makefile in a specific way?

@nayres Checkout this Makefile which is already doing what we're after
- https://github.com/hashicorp/terraform-plugin-framework/blob/main/Makefile

This is an example of how to call Makefile targets nested under `./website/`

bflad (approved):
Website loads okay with `make website`, `make website/build-local`, and `make website/local` -- I think we need the terraform-website bits setup for wiring /terraform/plugin/testing URLs to verify this local content unless there's another way.

PR #96, opened 2023-03-09
helper/resource: Deprecate non-testing constants, functions, and types

bflad:
The initial creation of the `terraform-plugin-testing` Go module was a migration of some existing packages within the `terraform-plugin-sdk` Go module. The `helper/resource` package itself contained multiple abstractions, some related to testing resources (the parts we want in this Go module) and some related to developing resources (the parts we do not want in this Go module). Version 1.0.0 with compatibility promises shipped before this extraneous code was pruned, so it is being deprecated in adherence to our versioning policy. It will not be included in any future major version of this Go module.

Each constant, function, method, and type includes a Go documentation deprecation comment with an explicit action for the developer. This is generally either copying the code or continuing to use the `terraform-plugin-sdk` Go module for effected usage.


bendbennett (approved):
LGTM 👍

bflad:
Yeah, that's definitely a good call.

bendbennett (approved):
LGTM 🚀

PR #102, opened 2023-03-16
Fix goreleaser deprecated `build` field

austinvalle:
Goreleaser deprecated the `build` field as it was previously not a documented field. Replaced with the equivalent `builds` array per deprecation guide: https://goreleaser.com/deprecations/#build

Failures seen in:
- https://github.com/hashicorp/terraform-plugin-testing/pull/101
- https://github.com/hashicorp/terraform-plugin-testing/pull/100


bflad (approved):
Looks good to me 🚀

PR #130, opened 2023-06-02
[COMPLIANCE] Add copyright and license headers

austinvalle:
Closes #44  

This PR introduces:
- `make generate` command that will auto-generate all copyright headers for this repo (hooks into `go generate`)
- GH workflow that will error if copyright headers aren't present or if license is malformed
- Addition to the `CONTRIBUTING.md` on how to run license autogenerating commands
- Add new `tools` module and add to dependabot


.copywrite.hcl
```diff
@@ -1,6 +1,27 @@
 schema_version = 1
 
 project {
-  license = "MPL-2.0"
+  license        = "MPL-2.0"
   copyright_year = 2014
```
austinvalle, line 5: An interesting discrepancy I'm guessing because this repo was ported over from plugin SDK's repo. When running `copywrite init` it sets the year to `2021`. Looking at the tool's logic, it is looking at [repo creation date](https://github.com/hashicorp/copywrite/blob/f7e172166ed22b27837654c8c349c607f975d60f/cmd/init.go#L82), not first commit date.

Not sure which is the correct one to use (all my other PRs have been using the `copywrite init` setup)

bendbennett: I would have thought the correct year was the year the repo was created.

bendbennett (approved):
LGTM 🎸

.copywrite.hcl
austinvalle: I double checked the RFC and you're correct. It says repo/project creation date, no comments about commit.

I will adjust to 2021

austinvalle: OOB we decided to preserve the original license year 👍🏻

PR #141, opened 2023-06-14
TestStep.ImportBlock: testing for import blocks

kmoe:
Add a new sub-mode of `TestStep.ImportState` for testing with the new [Terraform v1.5 `import` block feature](https://developer.hashicorp.com/terraform/language/import). The aim of this PR is to allow provider developers to add tests for `import` blocks with as little work as possible.

The test framework can generate an `import` block from the parameters already required for an ImportState test, e.g.
```
import {
  to = test_resource.foo
  id = "123"
}
```

Provider developers can optionally write their own import block and supply it with `TestStep.ImportBlockConfig`.

Other functionality for finding the import ID and verifying the import state is shared with existing ImportState testing.

## Example usage

An existing ImportState `TestStep` can be converted into import block tests with the addition of `ImportBlock: true`, e.g.

```
		Steps: []resource.TestStep{
			{
				Config: `resource "test_resource" "foo" {}`,
			},
			{
				ResourceName:      resourceName,
				ImportState:       true,
				ImportBlock:       true,
				ImportStateVerify: true,
			},
		},
```

## Design questions

1. Should `ImportBlock` be a separate `TestStep` mode entirely, with new `TestStep.ImportBlock*` fields? 

## To do

 - [ ] Deduplicate code between modes
 - [ ] Handle `TestStep.ImportResourceGenerateConfig` (for testing `terraform plan -generate-config-out`)
 - [ ] Docs (where do these live?)

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -560,11 +560,24 @@ type TestStep struct {
 	//---------------------------------------------------------------
 	// ImportState testing
 	//---------------------------------------------------------------
+	// Terraform has two workflows for importing resources: the import CLI
+	// command, which writes directly to state, and the import block in HCL,
+	// which imports to state via the normal plan and apply workflow.
 
 	// ImportState, if true, will test the functionality of ImportState
 	// by importing the resource with ResourceName (must be set) and the
 	// ID of that resource.
+	// By default, the "terraform import" command will be run. To test import
+	// block functionality instead, set ImportBlock to true.
 	ImportState bool
+
+	// ImportBlock, if true, enables a sub-mode of ImportState testing. In this
+	// mode, an import block is added to the config, and plan and apply are run.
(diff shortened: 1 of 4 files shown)


bflad:
Hi @kmoe 👋 Thank you for submitting this. Is there any urgency to this feature? We have some design inflection points coming up this quarter to get away from forever growing `TestStep` type, so it might be best suited to discuss this feature within that context.

bbasata:
Oh wow, this looks useful 😃 I'm doing some exploration in #431 inspired by this.

bbasata:
> Oh wow, this looks useful 😃 I'm doing some exploration in https://github.com/hashicorp/terraform-plugin-testing/pull/431 inspired by this.

We're shipping plannable import testing in v1.13.x ✨

PR #143, opened 2023-06-21
Logs successful and unsuccessful sweeper names instead of writing to stdout

gdavison:
Successful and unsuccessful sweeper names are currently written to stdout instead of the logger, while other output goes to the logger. This PR updates the output destination to be the logger


austinvalle (approved):
Thanks for the fix! 🚀

Issue #150, opened 2023-07-13
Add Option to use Terraform Configuration in TestStep 


bendbennett:
### terraform-plugin-testing version

```
v1.3.0
```

### Use cases

An individual acceptance test makes use of a [TestStep](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep) which includes a [Config](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep.Config) field containing a string representation of [Terraform configuration](https://developer.hashicorp.com/terraform/language), supplied as either [HCL](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md) or [JSON](https://developer.hashicorp.com/terraform/language/syntax/json). 

One of the consequences of the [Config](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep.Config) field using a string is that provider developers can decide on how they pass in Terraform configuration to the tests, for example:

- Usage of [raw string syntax](https://go.dev/ref/spec#String_literals) (backticks in Go).
- Calling Go standard library functions, such as [fmt.Sprintf()](https://pkg.go.dev/fmt#Sprintf), to dynamically generate the string.
- Usage of  a Go [constant](https://go.dev/ref/spec#Constants) or [variable](https://go.dev/ref/spec#Variables), with a value from either above form.
- Creation a Go [function](https://go.dev/ref/spec#Function_declarations) which returns a string (and may or may not take input parameters to make it more dynamic).
- Manually implement a Go [function](https://go.dev/ref/spec#Function_declarations) which can read an existing Terraform configuration file.

In all of the above cases except for the last, this can result in varying degrees of consistency in the provider codebase with respect to how Terraform configuration is created and supplied to the tests. Furthermore the usage of a Go string for Terraform configuration means that existing Terraform ecosystem tooling cannot be leveraged. This includes, but is not limited to:

- [terraform fmt](https://developer.hashicorp.com/terraform/cli/commands/fmt)
- [terraform validate](https://developer.hashicorp.com/terraform/cli/commands/validate)
- Editor code syntax highlighting.
- Tooling which can rewrite configurations.
- Cost estimation of running all acceptance tests.
- Collection of data regarding which tests cover a specific resource or data source.

### Attempted solutions

[...]

PR #153, opened 2023-07-17
Add Option to use Terraform Configuration in TestStep

bendbennett:
Closes: #150 

This PR adds the following fields to `TestStep`:

- `ConfigDirectory`
- `ConfigFile`
- `ConfigVariable`

### ConfigDirectory
Allows the specification of a path to a directory containing Terraform configuration files. All files in the directory are copied to the working directory used during the execution of an acceptance test `TestStep`.

### ConfigFile
Allows the specification of a path to a file containing Terraform configuration. The file is copied to the working directory used during the execution of an acceptance test `TestStep`.

### ConfigVariables
Allows defining Terraform variables for use in conjunction with Terraform configuration used during the execution of an acceptance test `TestStep`.


internal/teststep/config.go
```diff
@@ -0,0 +1,341 @@
...
+	testCaseProviderConfigFileName = "test_case_provider_config.tf"
+	testStepProviderConfigFileName = "test_step_provider_config.tf"
+)
+
+var (
+	providerConfigBlockRegex  = regexp.MustCompile(`provider "?[a-zA-Z0-9_-]+"? {`)
+	terraformConfigBlockRegex = regexp.MustCompile(`terraform {`)
+)
+
+type Config interface {
```
bendbennett, line 26: `HasDirectory()`, `HasRaw()`, `Raw()` and `WriteDirectory()` can all be removed and replaced with `Write(dest string)` once the logic for writing the raw configuration derived from `TestStep.Config` has been moved out of `*WorkingDir.SetConfig()`.

internal/teststep/config.go
```diff
@@ -0,0 +1,310 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+	}
+	return nil
+}
+
+func copyFile(path string, dstPath string) error {
```
bflad, line 188: [`internal/plugintest`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/internal/plugintest) has some helpers for this sorta thing -- if it makes more sense, maybe that functionality can split into its own internal file handling package to help prevent duplicating logic, but totally optional of course.

internal/teststep/config.go
```diff
@@ -0,0 +1,310 @@
...
+	return c.directory != ""
+}
+
+// HasProviderBlock returns true if the Config has declared a provider
+// configuration block, e.g. provider "examplecloud" {...}
+func (c configuration) HasProviderBlock(ctx context.Context) (bool, error) {
+	switch {
+	case c.hasRaw(ctx):
+		return providerConfigBlockRegex.MatchString(c.raw), nil
+	case c.hasDirectory(ctx):
```
bflad, line 97: 💡 One thing you could do here to avoid checking each case repetitively introduce two separate `configuration` implementations -- `ConfigurationDirectory` and `ConfigurationString`

config/directory.go
```diff
@@ -0,0 +1,59 @@
...
+
+// TestStepConfigFunc is the callback type used with acceptance tests to
+// specify a string which identifies a directory containing Terraform
+// configuration files.
+type TestStepConfigFunc func(TestStepConfigRequest) string
+
+// TestStepConfigRequest defines the request supplied to types
+// implementing TestStepConfigFunc.
+type TestStepConfigRequest struct {
+	StepNumber int
```
bflad, line 20: I think we may be able to add `TestName string` as well since the "front door" `resource.Test()` and friends should already be accepting a `testing.T`. 👍 I'll put a reciprocal comment on the function as well.

config/directory.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TestNameDirectory returns the name of the test when TestStepConfigFunc
+// is executed. This facilitates a convention of naming directories
+// containing Terraform configuration files with the name of the test.
+func TestNameDirectory(t *testing.T) func(TestStepConfigRequest) string { //nolint:paralleltest //Not a test
+	return func(_ TestStepConfigRequest) string {
+		return t.Name()
+	}
+}
```
bflad, line 48: If we can thread `t.Name()` through to `TestStepConfigRequest` like I think we can, we should be able to drop the `testing.T` here. It'd also be awesome to show an example of what happens from the developer perspective in the Go documentation. 👍 

One 🌶️  take that I have is that I think we should encourage folks to use the Go conventional `testdata` directory as well, which I'm showing in the suggestion here. Reference: https://pkg.go.dev/cmd/go#hdr-Test_packages / https://github.com/golang/go/issues/14715 / https://go-review.googlesource.com/c/go/+/23353/3/src/cmd/go/test.go

```suggestion
// TestNameDirectory returns the name of the test when TestStepConfigFunc
// is executed. This facilitates a convention of naming directories
// containing Terraform configuration files with the name of the test.
//
// For example, given test code:
//
//    func TestExampleCloudThing_basic(t *testing.T) {
//        resource.Test(t, resource.TestCase{
//            Steps: []resource.TestStep{
//                {
//                    ConfigDirectory: config.TestNameDirectory(),
//                },
//            },
//        })
//    }
//
// The testing configurations will be expected in the
// testdata/TestExampleCloudThing_basic/ directory.
func TestNameDirectory() func(TestStepConfigRequest) string {
	return func(req TestStepConfigRequest) string {
		return filepath.Join("testdata", req.TestName)
	}
}
```

bflad (commented):
Nice work so far, @bendbennett! I left some initial thoughts here. I really like where you are going with the new configuration handling interface and methods since that will help clean up the testing logic implementation.

config/variable.go
```diff
@@ -0,0 +1,319 @@
...
+		buf.Write([]byte(fmt.Sprintf("%q: ", k)))
+		buf.Write(j)
+		buf.Write([]byte(","))
+	}
+
+	b := bytes.TrimRight(buf.Bytes(), ",")
+
+	buf = bytes.NewBuffer(b)
+
+	buf.Write([]byte(`}`))
```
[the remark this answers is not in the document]
bendbennett: Good call. Now using `json.Marshal(v)` directly.

config/variable.go
```diff
@@ -0,0 +1,319 @@
...
+	b, err := json.Marshal(v.value)
+
+	if err != nil {
+		innerErr := err
+
+		for errors.Unwrap(innerErr) != nil {
+			innerErr = errors.Unwrap(err)
+		}
+
+		return nil, innerErr
```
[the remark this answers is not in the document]
bendbennett: Have added a comment. Happy to expand/reduce etc.

config/variable.go
```diff
@@ -0,0 +1,319 @@
...
+func ObjectVariable(value map[string]Variable) objectVariable {
+	return objectVariable{
+		value: value,
+	}
+}
+
+var _ Variable = numberVariable{}
+
+type number interface {
+	constraints.Float | constraints.Integer | string
```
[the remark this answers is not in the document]
bendbennett: Have implemented `FloatVariable` and `IntegerVariable`. Have punted on `NumberVariable` for now but happy to add.

config_directory_test/config_directory_test.go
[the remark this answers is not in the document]
bendbennett: I believe this is now resolved via the following code change for detection of symlinks:

```go
if dirEntry.IsDir() || fi.Mode()&fs.ModeSymlink == fs.ModeSymlink {
```

internal/teststep/config.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+type configuration struct {
+	directory string
+	raw       string
+}
+
+type ConfigurationRequest struct {
+	Directory string
+	Raw       string
```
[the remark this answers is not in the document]
bendbennett: Done 👍

internal/teststep/config.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func Configuration(req ConfigurationRequest) (configuration, error) {
+	var populatedConfig []string
```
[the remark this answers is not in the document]
bendbennett: Have added a `ConfigurationRequest.Validate()` method as per your suggestion.

internal/teststep/config.go
bendbennett: I've left this as is for now if that's ok? Once these changes are merged I'll add an issue to consider refactoring.

internal/teststep/config.go
bendbennett: I've added `configurationDirectory` and `configurationString` types that implement the `Config` interface. The `Configuration()` constructor function now returns `Config` interface to accommodate the various types we will need (e.g., `configurationDirectory`, `configurationString`, `configurationFile`). Owing to the switch to returning an interface from the `Configuration()` constructor function, nil checks have been added.

config/directory.go
bendbennett: Added.

internal/teststep/directory_test.go
```diff
@@ -0,0 +1,576 @@
...
+			},
+			expected: false,
+		},
+		"provider-block-quoted-with-attributes": {
+			configDirectory: configurationDirectory{
+				directory: "testdata/provider_block_quoted_with_attributes",
+			},
+			expected: true,
+		},
+		"provider-block-unquoted-with-attributes": {
```
austinvalle, line 60: TIL this is valid 😆

website/docs/plugin/testing/acceptance-tests/configuration.mdx
```diff
@@ -0,0 +1,277 @@
+---
```
austinvalle, line 1: Do we want to callout this documentation page on the existing `/terraform/plugin/testing/acceptance-tests/teststep#test-modes` page when talking about `Lifecycle` mode? 

All the examples in that file use the raw `Config`, but that's probably fine

website/docs/plugin/testing/acceptance-tests/configuration.mdx
```diff
@@ -0,0 +1,277 @@
...
+
+The `TestStep.ConfigDirectory` field accepts a [TestStepConfigFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigFunc) which is a function that accepts a [TestStepConfigRequest](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigRequest) and returns a string containing a path to a directory containing Terraform configuration files. The path can be a relative or absolute path.
+
+There are helper methods available for generating a [TestStepConfigFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigFunc) including:
+
+* [StaticDirectory(directory string)](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#StaticDirectory)
+* [TestNameDirectory()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestNameDirectory)
+* [TestStepDirectory()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepDirectory)
+
+~> **Note**: `ExternalProviders` cannot be specified when using ConfigDirectory.
```
austinvalle, line 56: Is the intention for developers to use the `required_providers` instead? Might be worth mentioning this doc: https://developer.hashicorp.com/terraform/language/providers/requirements#requiring-providers

website/docs/plugin/testing/acceptance-tests/configuration.mdx
```diff
@@ -0,0 +1,277 @@
...
+
+The `TestStep.ConfigFile` field accepts a [TestStepConfigFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigFunc) which is a function that accepts a [TestStepConfigRequest](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigRequest) and returns a string containing a path to a file containing Terraform configuration. The path can be a relative or absolute path.
+
+There are helper methods available for generating a [TestStepConfigFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepConfigFunc) including:
+
+* [StaticFile(file string)](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#StaticFile)
+* [TestNameFile(file string)](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestNameFile)
+* [TestStepFile(file string)](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/config#TestStepFile)
+
+~> **Note**: `ExternalProviders` cannot be specified when using `ConfigFile`.
```
austinvalle, line 139: Same note as above: https://developer.hashicorp.com/terraform/language/providers/requirements#requiring-providers

austinvalle (approved):
So much great work Ben! 🚀  - Couple notes and one thought here:

- With the amount of file/path specific tests we have now in `plugin-testing`, should we update our CI to run on other platforms as well (mac/windows)?

helper/resource/teststep_providers_test.go
```diff
@@ -2349,6 +2247,848 @@ func TestTest_TestStep_ProviderFactories_Import_External_With_Data_Source(t *tes
...
+	})
+}
+
+func TestTest_ConfigDirectory_TestStepDirectory(t *testing.T) {
+	t.Parallel()
+
+	Test(t, TestCase{
+		Steps: []TestStep{
+			{
+				ConfigDirectory: config.TestStepDirectory(),
```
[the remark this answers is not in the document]
bendbennett: Have added a test to demonstrate that the TestStep number is iterating and using config from the appropriately numbered directory.

internal/plugintest/util.go
```diff
@@ -125,7 +126,17 @@ func CopyDir(src, dest, baseDirName string) error {
...
+
+		if err != nil {
+			return fmt.Errorf("unable to get dir entry info: %w", err)
+		}
+
+		if dirEntry.IsDir() || fi.Mode()&fs.ModeSymlink == fs.ModeSymlink {
+			if err = CopyDir(srcFilepath, destFilepath, baseDirName); err != nil {
+				return fmt.Errorf("unable to copy directory: %w", err)
+			}
+		} else if fi.Mode() == fs.ModeSymlink {
```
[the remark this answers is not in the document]
bendbennett: Good spot! The `else if` is redundant and actually wouldn't work as shown. Have removed.

internal/teststep/directory_test.go
bendbennett: Me too 😆

website/docs/plugin/testing/acceptance-tests/configuration.mdx
bendbennett: Good point. Have added a sentence to make this clearer and used the link you supplied. Thanks

website/docs/plugin/testing/acceptance-tests/configuration.mdx
bendbennett: Good point. Have added a sentence to make this clearer and used the link you supplied. Thanks

bendbennett:
> * With the amount of file/path specific tests we have now in `plugin-testing`, should we update our CI to run on other platforms as well (mac/windows)?

We could certainly consider this. Couple of observations when playing around with this:

- Updating the OS matrix in `ci-go.yml` to use windows-latest and macos-latest results in the error below when running on Windows. This can be handled by changing the running of the tests to something like `go test ./... -v -cover`:

```
no required module provides package .out; to add it: .........
```

- Next up, there's an [error](https://github.com/hashicorp/terraform-plugin-testing/actions/runs/5712009551/job/15474925149) generated for the `TestTest_TestStep_ProviderFactories_CopyWorkingDir_EachTestStep` test that's specific to MacOS 12. This can be handled by using `macos-13` in the testing matrix, but will require that we circle back round and update to `macos-latest` once the default for MacOS for GitHub actions has been bumped to 13.

- Finally, the tests that purposely fail to find a directory/file, for instance, `TestConfigurationFile_HasProviderBlock` "not-file", will need to be updated as the error message on Windows differs from Ubuntu and MacOS (e.g., `The system cannot find the file specified` versus `no such file or directory`).

I'm happy to make these changes if everyone is on board with the above or some variant thereof.

bflad (commented):
Drive-by question (sorry I'm having trouble getting time to review this properly): Is there something handling the removal of configuration files not present when there are multiple `ConfigDirectory` steps?

e.g. if first test step has `ConfigDirectory` with `a.tf` and `b.tf.json` and second test step has `ConfigDirectory` has `c.tf` and `d.tf.json`, the working directory should only have `c.tf` and `d.tf.json` during the second test step.

helper/resource/teststep_providers_test.go
```diff
@@ -2333,6 +2334,25 @@ func TestTest_ConfigDirectory_TestStepDirectory(t *testing.T) {
 	})
 }
 
+// TestTest_ConfigDirectory_TestStepDirectory_StepNotHardcoded uses a multistep test
+// to prove that the test step number is not hardcoded
```
austinvalle, line 2338: 🤣 🤣

austinvalle:
> > * With the amount of file/path specific tests we have now in `plugin-testing`, should we update our CI to run on other platforms as well (mac/windows)?
> 
> We could certainly consider this. Couple of observations when playing around with this:
> 
> * Updating the OS matrix in `ci-go.yml` to use windows-latest and macos-latest results in the error below when running `go test -coverprofile=coverage.out ./...` on Windows. This can be handled by changing the running of the tests to something like `go test ./... -v -cover`:
> 
> ```
> no required module provides package .out; to add it: .........
> ```
> 
> * Next up, there's an [error](https://github.com/hashicorp/terraform-plugin-testing/actions/runs/5712009551/job/15474925149) generated for the `TestTest_TestStep_ProviderFactories_CopyWorkingDir_EachTestStep` test that's specific to MacOS 12. This can be handled by using `macos-13` in the testing matrix, but will require that we circle back round and update to `macos-latest` once the default MacOS for GitHub actions has been bumped to 13.
> * Finally, the tests that purposely fail to find a directory/file, for instance, `TestConfigurationFile_HasProviderBlock` "not-file", will need to be updated as the error message on Windows differs from Ubuntu and MacOS (e.g., `The system cannot find the file specified` versus `no such file or directory`).
> 
> I'm happy to make these changes if everyone is on board with the above or some variant thereof.

Hmm 🤔 , I don't want it to be too big of a pain... Maybe we just skip running them for different OS's for now? If we start running into bug reports that are OS specific we can revisit. I'd imagine the amount of users running into Window specific bugs is limited for Go modules like this anyways 🤷🏻

bendbennett:
> Drive-by question (sorry I'm having trouble getting time to review this properly): Is there something handling the removal of configuration files not present when there are multiple `ConfigDirectory` steps?
> 
> e.g. if first test step has `ConfigDirectory` with `a.tf` and `b.tf.json` and second test step has `ConfigDirectory` has `c.tf` and `d.tf.json`, the working directory should only have `c.tf` and `d.tf.json` during the second test step.

Good point! Have added something to handle removal of config and variables files from previous test step prior to running the current test step along with some test coverage.

config/directory.go
```diff
@@ -0,0 +1,74 @@
...
+
+// TestStepConfigFunc is the callback type used with acceptance tests to
+// specify a string which identifies a directory containing Terraform
+// configuration files.
+type TestStepConfigFunc func(TestStepConfigRequest) string
+
+// TestStepConfigRequest defines the request supplied to types
+// implementing TestStepConfigFunc.
+type TestStepConfigRequest struct {
+	StepNumber int
```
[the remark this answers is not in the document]
bendbennett: Have updated the docs to this effect.

config/directory.go
```diff
@@ -0,0 +1,74 @@
...
+		return filepath.Join("testdata", req.TestName)
+	}
+}
+
+// TestStepDirectory returns the name of the test suffixed with an
+// OS specific separator and the test step number. This facilitates
+// a convention of naming directories containing Terraform
+// configuration files with the test step number and nesting of
+// these files within a directory with the same name as the test.
+func TestStepDirectory() func(TestStepConfigRequest) string { //nolint:paralleltest //Not a test
```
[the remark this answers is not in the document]
bendbennett: Have updated the docs to this effect.

internal/plugintest/working_dir.go
```diff
@@ -82,29 +82,71 @@ func (wd *WorkingDir) GetHelper() *Helper {
...
+
+	fi, err := d.Readdir(-1)
+
+	if err != nil {
+		return err
+	}
+
+	for _, file := range fi {
+		if file.Mode().IsRegular() {
+			if filepath.Ext(file.Name()) == ".tf" || filepath.Ext(file.Name()) == ".json" {
```
bflad, line 103: Open question: Not sure if we should or shouldn't be concerned about non-Terraform JSON files here.

bflad (approved):
Overall looks good to me 🚀

internal/plugintest/working_dir.go
bendbennett: I _think_ that the only `.json` files within the `working_dir` will be those that are copied when a `.json` file, or a directory containing `.json` files is specified for `TestStep.ConfigFile`, or`TestStep.ConfigDirectory`, respectively. So I believe removing `.json` files will be ok from this perspective.

Do you know any way in which a non-Terraform JSON file might end-up in the working directory, and if so, whether removing it would be likely to cause issues?

bflad: Hmm, I think you're right to a certain degree. Maybe I was thinking of the opposite problem then -- let's say we're using `ConfigDirectory` across multiple steps and I have:

```
# step 1 directory
|- file.whateverextension
 \ main.tf

# step 2 directory
\ main.tf
```

We should be preserving things necessary between steps in the working directory such as the `terraform.tfstate`, but removing the rest of the working directory files/directories. The configuration copying would overwrite any lingering files between steps, so in that regard its okay, but I think it would issue if the configuration was using something like `fileset()` instead of referencing thing with `file()` directory _and_ expecting those lingering files to be removed between steps. Let's treat this as something to follow up with rather than holding this up further though!

bflad (approved):

internal/plugintest/working_dir.go
bendbennett: Thanks for the additional context. Have created a follow-up issue - [Consider removal of all files copied from ConfigDirectory between sequential test steps](https://github.com/hashicorp/terraform-plugin-testing/issues/172#top)

Issue #151, opened 2023-07-13
Migrate Testing Provider Code to terraform-plugin-go


bflad:
### Description

In order to appropriately (unit) test the functionality offered to provider developers with `helper/resource.TestCase` and friends, it generally requires the need for real provider implementations. Currently, some of the tests utilize the lingering direct dependency on terraform-plugin-sdk. For example:

```go
{
	Config: `resource "null_resource" "test" {}`,
	ProviderFactories: map[string]func() (*schema.Provider, error){
		"null": func() (*schema.Provider, error) { //nolint:unparam // required signature
			return &schema.Provider{
				ResourcesMap: map[string]*schema.Resource{
					"null_resource": {
						CreateContext: func(_ context.Context, d *schema.ResourceData, _ interface{}) diag.Diagnostics {
							d.SetId("test")
							return nil
						},
						DeleteContext: func(_ context.Context, _ *schema.ResourceData, _ interface{}) diag.Diagnostics {
							return nil
						},
						ReadContext: func(_ context.Context, _ *schema.ResourceData, _ interface{}) diag.Diagnostics {
							return nil
						},
						Schema: map[string]*schema.Schema{
							"triggers": {
								Elem:     &schema.Schema{Type: schema.TypeString},
								ForceNew: true,
								Optional: true,
								Type:     schema.TypeMap,
							},
						},
					},
				},
			}, nil
		},
	},
},
```

Other tests rely on downloading real providers, such as `hashicorp/random` that happen to include the desired functionality for testing.

A long term goal of this Go module would be to remove that direct dependency, which is only based on supporting the legacy `Providers` and `ProviderFactories` fields which work with the `terraform-plugin-sdk/helper/schema.Provider` type.

### Proposal

The summary of this proposal is:

- Introduce a relatively straightforward terraform-plugin-go based "provider SDK" for creating testing providers in this codebase
- Remove any testing provider code that is directly terraform-plugin-sdk based.
- Real, externally hosted providers, such as `hashicorp/random` etc., can and should still be used for testing `ExternalProviders` field functionality, but not in other cases.

Some additional notes/assumptions:

- Provider developers should always want to test in-development provider Go code, rather than being forced to host/publish them elsewhere first.
- The terraform-plugin-go dependency should always exist in this Go module as long as that Go module is the one that contains the provider protocol implementation in Go or at least the lowest-level provider server implementation.
[...]

PR #164, opened 2023-08-21
helper/resource: Remove data source and resource `id` attribute requirement

### LINKED ISSUE #84: helper/resource: Try to Remove id Attribute Requirement
### SDK version

```
v2.23.0
```

### Use-cases

Providers written using terraform-plugin-go and terraform-plugin-framework do not require an `id` schema attribute, nor is it required by Terraform or the protocol. There are legacy reasons why terraform-plugin-sdk requires the `id` attribute but maybe that can be removed from the testing framework. This would prevent those providers from needing to declare placeholder `id` attributes in their schemas to prevent this type of error:

```
testing_new_config.go:111: no "id" found in attributes
testing_new.go:53: no "id" found in attributes
```

Then needing to discover the callout in the documentation.

### Attempted Solutions

Placeholder `id` schema attributes.

### Proposal

See if we cannot convince the legacy Terraform resource state handling underlying helper/resource that it is okay that we do not have an `id` schema attribute.

### References

- https://www.terraform.io/plugin/framework/acctests#implement-id-attribute

bflad:
Closes #84

The resource instance state identifier was a Terraform versions 0.11 and earlier concept which helped core and the then SDK determine if the resource should be removed and as an identifier value in the human readable output. This concept unfortunately carried over to the testing logic when the testing logic was mostly changed to use the public, machine-readable JSON interface with Terraform, rather than reusing prior internal logic from Terraform. Using the `id` attribute value for this identifier was the default implementation and therefore those older versions of Terraform required the attribute. This is no longer necessary after Terraform versions 0.12 and later. This testing logic only supports Terraform version 0.12.26 and later.

The remaining resource instance identifier and `id` attribute usage in the testing logic was:

- When the `TestStep` type `ImportStateVerify` field is enabled, the instance state identifier between prior resource state and import resource state are compared to find the correct resource.
- When the `TestStep` type `ImportStateIdFunc` and `ImportStateId` fields are empty as the import ID for calling `terraform import` command.
- Other various `terraform` package references

For the first case, a new `TestStep` type `ImportStateVerifyIdentiferAttribute` field is added, which defaults to the prior `id` attribute. This preserves the existing behavior while enabling developers to choose a different identifier attribute if omitted from the resource so they can still use the `ImportStateVerify` functionality.

For the second case, developers can use the `ImportStateId*` fields to choose/create the correct import identifier value.

For the final cases, no action is needed as the `terraform` package is not intended for external usage and any usage of this identifier is mainly for equality and printing debugging information which is not accessed by the testing logic. These functions will be marked with Go documentation deprecation comments in a subsequent change.

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20230820-195300.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20230820-195300.yaml
@@ -1,0 +1,7 @@
+kind: ENHANCEMENTS
+body: 'helper/resource: Added `TestStep` type `ImportStateVerifyIdentifierAttribute`
+  field, which can override the default `id` attribute used for matching prior
+  resource state with imported resource state'
+time: 2023-08-20T19:53:00.488517-04:00
+custom:
+  Issue: "84"
--- a/.changes/unreleased/FEATURES-20230820-195624.yaml
+++ b/.changes/unreleased/FEATURES-20230820-195624.yaml
@@ -1,0 +1,6 @@
+kind: FEATURES
+body: 'helper/resource: Removed data resource and managed resource `id` attribute
+  requirement'
+time: 2023-08-20T19:56:24.356163-04:00
+custom:
+  Issue: "84"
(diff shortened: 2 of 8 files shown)


bflad:
Draft note: Testing this will require some form of #151 so we don't need to rely on external providers that are 1) not written in terraform-plugin-sdk code and 2) do not have an `id` attribute

austinvalle (approved):
lgtm 🚀

PR #166, opened 2023-08-22
internal/testing: Initial testsdk and testprovider packages

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/151

Creates an internal terraform-plugin-go based SDK and declarative provider for unit testing for within this Go module. The new `helper/resource` tests should be equivalent to the existing terraform-plugin-sdk based `TestTest_TestStep_ExternalProviders_To_ProviderFactories` and `TestTest_TestStep_ProviderFactories_To_ExternalProviders` tests.


austinvalle (approved):
LGTM 🚀

PR #170, opened 2023-08-28
helper/resource: Refactor various unit testing to use inline providers

### LINKED ISSUE #151: Migrate Testing Provider Code to terraform-plugin-go

bflad:
Closes #151

This will reduce CI time and dependency on networking/registry calls. Other unit testing that can be refactored can be done when the terraform-plugin-sdk dependency is being removed.

### DIFF
--- a/helper/resource/testing_new_config_test.go
+++ b/helper/resource/testing_new_config_test.go
@@ -8,24 +8,52 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testprovider"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/providerserver"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 )
 
 func TestTest_TestStep_ExpectError_NewConfig(t *testing.T) {
 	t.Parallel()
 
-	Test(t, TestCase{
-		ExternalProviders: map[string]ExternalProvider{
-			"random": {
-				Source:            "registry.terraform.io/hashicorp/random",
-				VersionConstraint: "3.4.3",
-			},
+	UnitTest(t, TestCase{
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"test": providerserver.NewProviderServer(testprovider.Provider{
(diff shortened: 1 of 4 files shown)


austinvalle (approved):
lgtm

PR #171, opened 2023-08-28
terraform: Add deprecation Go documentation comments

### LINKED ISSUE #165: Deprecate Unreferenced terraform Package Functionality

```
v1.4.0
```

### Use cases

The `terraform` package in this Go module contains legacy, unused functionality that has been continually passed down since the code was originally in the core codebase. This functionality is not intended for external consumption, but it has been passed down in its original exported API form. To help prevent unexpected consumption of this functionality, it should be marked as deprecated for removal. Any future major version of this Go module _will_ remove the _entire_ `terraform` package.

### Proposal

Add Go documentation deprecation comments to all `terraform` package functionality that is _not_ referenced by `helper/resource` currently.

bflad:
Closes #165

Except in certain known provider testing use cases without replacement (yet), this deprecates various `terraform` package functionality. The `terraform` package contains legacy Terraform core logic which has been copied to terraform-plugin-sdk and terraform-plugin-testing over the years and continually exported due to the complexity of rewriting developer functionality using the machine-readable interfaces for Terraform, such as JSON state. Much of this now-deprecated `terraform` package logic should have been omitted when this Go module was created, but that step was missed, so it is left as-is with deprecation notices following the Go module versioning guidelines. In reality, the entire `terraform` package will be removed in a future major version, however this change pragmatically leaves certain functionality without deprecation notices for now until  replacement functionality is available so developers will not need to silence linting tools without it being actionable yet.

If there is a valid use case for this functionality, developers should create a GitHub issue for tracking.

### DIFF
--- a/helper/resource/state_shim.go
+++ b/helper/resource/state_shim.go
@@ -21,7 +21,7 @@ type shimmedState struct {
 }
 
 func shimStateFromJson(jsonState *tfjson.State) (*terraform.State, error) {
-	state := terraform.NewState()
+	state := terraform.NewState() //nolint:staticcheck // legacy usage
 	state.TFVersion = jsonState.TerraformVersion
 
 	if jsonState.Values == nil {
@@ -124,7 +124,7 @@ func (ss *shimmedState) shimStateModule(sm *tfjson.StateModule) error {
 		}
 	}
 
-	mod := ss.state.AddModule(path)
+	mod := ss.state.AddModule(path) //nolint:staticcheck // legacy usage
 	for _, res := range sm.Resources {
 		resourceState, err := shimResourceState(res)
 		if err != nil {
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -1548,7 +1548,7 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
 // TestCheckOutput checks an output in the Terraform configuration
 func TestCheckOutput(name, value string) TestCheckFunc {
 	return func(s *terraform.State) error {
-		ms := s.RootModule()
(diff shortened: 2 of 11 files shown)


austinvalle (approved):
lgtm 🚀

Do we typically add changelogs for adding deprecation notices? Or just wait until the actual deprecation occurs

bflad:
> Do we typically add changelogs for adding deprecation notices?

Sorry I should've mentioned that in the commit message.

The short answer is yes for intentionally designed Go packages, however I'm opting to _not_ include a changelog entry here because I think it'll be more confusing for developers, who might feel like they need to go check for usage when it likely affects only a very small (if any) portion of the population. Their usage externally would be unexpected as all of these are unintentionally part of the exported Go API having been just copied Terraform core "internal" code from some historical time and we likely would not support any changes to their logic since they fall outside the intention of the Go module being acceptance testing of Terraform Providers. If/when we do deprecate something like the `terraform.State` type, which is externally prevalent thanks to `helper/resource.TestCheckFunc`, that _definitely_ warrants a changelog entry (and maybe even website documentation and upgrade tooling) with its replacement. I tried to be careful and leave other somewhat prevalent functionality under the TestCheckFunc expected use case, such as `(terraform.State).RootModule()`, also without deprecation comments for now since there is no replacement yet. We can undo any "oops this is regularly called by providers implementing their own TestCheckFunc" by removing the deprecation comments.

We can certainly add a changelog entry in this case if folks disagree with my sentiment though. Happy to discuss more.

austinvalle:
> _however I'm opting to not include a changelog entry here because I think it'll be more confusing for developers_

This makes sense to me; changelog should be geared for consumers of the module and if a note would bring confusion omitting sounds like a good call. Thanks for the context 👍🏻

Issue #215, opened 2023-11-13
Add terraform block to config if not already present


bendbennett:
### terraform-plugin-testing version

```
v1.5.1
```

### Use cases

In order to support the testing of provider defined functions it is necessary to ensure that a `terraform {...}` block appears in the configuration that is being used. 

### Attempted solutions

N/A

### Proposal

Inspect the configuration that has been generated to determine whether a `terraform {...}` block is present, and if not, prefix the configuration with a `terraform {}` block.

PR #216, opened 2023-11-13
Add terraform block to config when not already present

bendbennett:
Closes: #215

Dependency: #228  

### Considerations/Questions

- Currently, there is no handling to add required provider entries into the terraform block if `TestStep.Config` contains JSON. 
  - Does this need implementing?
- The entries in the terraform required providers block are not added prior to the initial call to [wd.Init](https://github.com/hashicorp/terraform-plugin-testing/pull/216/files#diff-5dff77f717877351ff8ecebd2f32cf8cfadbdecf71cae9ee923a5079083208d7R102), or in [testIDRefresh()](https://github.com/hashicorp/terraform-plugin-testing/pull/216/files#diff-5dff77f717877351ff8ecebd2f32cf8cfadbdecf71cae9ee923a5079083208d7R531) 
  - Interested to hear comments on whether this is viewed as ok. 
- The [providerConfigTestCase()](https://github.com/hashicorp/terraform-plugin-testing/pull/216/files#diff-1805fc800592b0f38323dc4bc35e7bea3ba7c3819bf9a4adfb6a38c4561910b4R146) removes entries from `testCase.Providers`, when an entry with the same key is found in `testCase.ProviderFactories`.  
  - This was required to avoid a panic generated by the presence of two entries with the same key in the terraform required providers block when running `TestTest_TestCase_Providers` (i.e., `TestTest_TestCase_Providers` results in `test` appearing as a key in both `testCase.Providers`, and `testCase.ProviderFactories`.


internal/teststep/config.go
```diff
@@ -126,6 +127,37 @@ func Configuration(req ConfigurationRequest) Config {
...
+
+// AddTerraformBlock inspects ConfigString, and prefixes
+// with an empty terraform block providing the configuration
+// string is not JSON, and a Terraform block is not already
+// present.
+func (c ConfigString) AddTerraformBlock() string {
+	if !c.IsJSON() && !c.HasTerraformBlock() {
+		str := strings.TrimLeft(string(c), "\n")
+
+		return fmt.Sprintf("terraform {}\n\n%s", str)
```
bflad, line 142: For this to work as expected, the `terraform` configuration in this case must include a correct `required_providers` definition to match the provider being tested, e.g. 

```terraform
terraform {
  required_providers {
    NAME = {
      source = "ADDRESS"
    }
  }
}
```

This is what future Terraform provider functionality will require in the configuration so Terraform explicitly knows which provider is being referenced in a configuration (rather than implicitly guessing any missing provider declaration is in the hashicorp namespace). Here's some internal background about this problem: https://github.com/hashicorp/terraform-proposals/issues/66

There is _somewhat_ of an example of this in `(TestStep).providerConfig()` but its only wired to the `ExternalProviders` information. This implementation needs to account for a lot more, unfortunately:

I believe the NAME values must contain all the map keys of all the following defined on the `TestCase`/`TestStep`:

* `ExternalProviders`
* `ProtoV5ProviderFactories`
* `ProtoV6ProviderFactories`

(We don't necessarily need to include `Providers`/`ProviderFactories` as terraform-plugin-sdk/v2 providers will never receive the future enhancements needing the required provider configuration.)

And the ADDRESS value _may_ also need to depend on things like the [`TF_ACC_PROVIDER_NAMESPACE` environment variable](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests#environment-variables). We might need to introduce some additional testing with non-hashicorp namespace providers and the `ProtoV{5,6}ProviderFactory` fields to verify whether that value must be consulted or not.

helper/resource/testcase_providers.go
```diff
@@ -21,6 +21,10 @@ func (c TestCase) providerConfig(_ context.Context, skipProviderBlock bool) stri
 	//      it does have a special purpose that wasn't being unit tested prior.
 	for name := range c.Providers {
 		providerBlocks.WriteString(fmt.Sprintf("provider %q {}\n", name))
+
+		requiredProviderBlocks.WriteString(fmt.Sprintf("    %s = {\n", name))
+
+		requiredProviderBlocks.WriteString("    }\n")
```
bflad, line 27: I think we can omit this handling here (at least for now) -- its odd to have empty required_provider configuration and honestly not sure if Terraform will always treat that as valid. 👍

helper/resource/teststep_providers.go
```diff
@@ -5,10 +5,15 @@ package resource
...
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
+
+	"github.com/hashicorp/go-version"
 )
 
+const tfBlockMinReqTFVersion = "1.6.0"
```
bflad, line 15: I think we discussed this out of band, but it might be good to set this to 1.0.0 with the reasoning that we'd eventually remove this version check entirely in the terraform-plugin-testing major version we drop protocol version 5 support. Another potentially good reason to do this now is that it won't potentially delay feedback for folks who may be already be testing on 1.x and upgrade their terraform-plugin-testing.

If you could also drop a quick code comment here about why this version check exists at all, that would be fantastic!

.changes/unreleased/ENHANCEMENTS-20231128-143901.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'helper/resource: Automatically add `required_providers` configuration to `TestStep.Config`
+  configuration when using Terraform 1.6.* and above'
```
bflad, line 3: Good consideration callout in the PR description about JSON, since we currently treat JSON as "you know what you're doing" and our future functionality testing documentation will always show Terraform configuration language, I think just putting a quick "Terraform language" before the word "configuration" here will suffice. Also if we do update the check to 1.0, that will need to be updated here as well. 😄

bflad (approved):
Overall, I think this is what we want! To answer some of your questions:

> Currently, there is no handling to add required provider entries into the terraform block if TestStep.Config contains JSON.
> Does this need implementing?

I'd say no for now -- we have treated JSON configuration as a "you know what you're doing", similar to the new configuration file/directory handling.

> The entries in the terraform required providers block are not added prior to the initial call to [wd.Init](https://github.com/hashicorp/terraform-plugin-testing/pull/216/files#diff-5dff77f717877351ff8ecebd2f32cf8cfadbdecf71cae9ee923a5079083208d7R102), or in [testIDRefresh()](https://github.com/hashicorp/terraform-plugin-testing/pull/216/files#diff-5dff77f717877351ff8ecebd2f32cf8cfadbdecf71cae9ee923a5079083208d7R531)

I think this is okay. Init only needs external providers (or whatever it currently gets). ID refresh is a very legacy testing feature that doesn't need to be touched. It honestly should be deprecated, but I forget why that hasn't already been done other than maybe previously thinking all the next major version deprecations would be done at once.

---

At some point we'll clean up all those gnarly helper/resource function signatures. 😬 😅

Issue #219, opened 2023-11-17
Add unknown and null plan checks for output values


bendbennett:
### terraform-plugin-testing version

```
v1.5.1
```

### Use cases

Provider defined functions are preprocessed during configuration loading and do not have their own state that gets saved. In order to be able to test the return value from a provider defined function it will be necessary to store the return value so that it can be checked during testing. One option for storage of these values is to use an [output value](https://developer.hashicorp.com/terraform/language/values/outputs). As the value of an output value, or a value nested within an output value can be null or unknown it would be beneficial to offer a simple way to be able to determine whether this was the case to ease testing whether this is the case.

### Proposal

The terraform-plugin-testing Go module already has plan checks available for evaluating whether an attribute at a specific address and path is unknown (refer to [Built-in Plan Checks](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks#built-in-plan-checks)). It is proposed that plan checks could be extended to allow for testing of unknown and null output values too.

PR #220, opened 2023-11-17
Implementing expect unknown and null output value plan checks

bendbennett:
Closes: #219


plancheck/expect_null_output_value.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if result != nil {
+		resp.Error = fmt.Errorf("attribute at path is not null")
+
+		return
+	}
+}
+
+// ExpectNullOutputValue returns a plan check that asserts that the specified attribute at the given resource has an unknown value.
```
SBGoods, line 63:
```suggestion
// ExpectNullOutputValue returns a plan check that asserts that the specified attribute at the given resource has a null value.
```

plancheck/expect_null_output_value.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// Due to implementation differences between the terraform-plugin-sdk and the terraform-plugin-framework, representation of unknown
```
SBGoods, line 65:
```suggestion
// Due to implementation differences between the terraform-plugin-sdk and the terraform-plugin-framework, representation of null
```

plancheck/expect_null_output_value.go
```diff
@@ -0,0 +1,73 @@
...
+// values may differ. For example, terraform-plugin-sdk based providers may have less precise representations of unknown values, such
```
SBGoods, line 66:
```suggestion
// values may differ. For example, terraform-plugin-sdk based providers may have less precise representations of null values, such
```

plancheck/expect_null_output_value.go
```diff
@@ -0,0 +1,73 @@
...
+// as marking whole maps as unknown rather than individual element values.
```
SBGoods, line 67:
```suggestion
// as marking whole maps as null rather than individual element values.
```

tfjsonpath/path.go
```diff
@@ -69,34 +82,38 @@ func (s Path) AtMapKey(key string) Path {
...
+
 			result, ok = mapObj[string(s)]
+
 			if !ok {
-				return nil, fmt.Errorf("path not found: specified key %s not found in map", string(s))
+				return nil, fmt.Errorf("path not found: specified key %s not found in map at %s", string(s), strings.Join(steps, "."))
 			}
 
 		case SliceStep:
+			steps = append(steps, fmt.Sprint(s))
```
SBGoods, line 107: Is there a use case for using `fmt.Sprint(s)` instead of `string(s)` here?

plancheck/expect_unknown_output_value.go
```diff
@@ -0,0 +1,110 @@
...
+// values may differ. For example, terraform-plugin-sdk based providers may have less precise representations of unknown values, such
+// as marking whole maps as unknown rather than individual element values.
+func ExpectUnknownOutputValue(params OutputValueParams) PlanCheck {
+	return expectUnknownOutputValue{
+		outputAddress: params.OutputAddress,
+		attributePath: params.AttributePath,
+	}
+}
+
+// OutputValueParams is used during the creation of a plan check for output values, and specifies
```
SBGoods, line 75: Is it better to move this information to the website documentation since it's directly exposed to provider developers?

SBGoods (commented):
Overall, this looks great! I've added minor comments mostly around documentation.

More generally, I think we should come up with common language to differentiate between attribute state and attribute output/value. Especially with `expectUnknownValue()` and `expectUnknownOutputValue()` where the language is a bit too similar. Maybe we can describe `expectUnknownValue()` and `expectSensitiveValue()` as "asserting that an attribute is being marked as unknown/sensitive."

website/docs/plugin/testing/acceptance-tests/plan-checks.mdx
```diff
@@ -21,13 +21,15 @@ A **plan check** is a test assertion that inspects the plan file at a specific p
...
-| [`plancheck.ExpectNonEmptyPlan()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck#ExpectNonEmptyPlan)                       | Asserts the entire plan contains at least one operation for apply.           |
-| [`plancheck.ExpectResourceAction(address, operation)`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck#ExpectResourceAction) | Asserts the given resource has the specified operation for apply.            |
-| [`plancheck.ExpectUnknownValue(address, path)`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck#ExpectUnknownValue)          | Asserts the specified attribute at the given resource has an unknown value.  |
-| [`plancheck.ExpectSensitiveValue(address, path)`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck#ExpectSensitiveValue)      | Asserts the specified attribute at the given resource has a sensitive value. |
+| Check                                                                                                                                                      | Description                                                                  |
```
bflad, line 24: It might be good to split up this table (and maybe the page) to be resource attribute checks versus output value checks to reduce potential confusion.

tfjsonpath/path.go
bendbennett: There's no use-case for `fmt.Sprint(s)`, have updated to `string(s)`.

plancheck/expect_unknown_output_value.go
```diff
@@ -0,0 +1,110 @@
...
+// Then the value cannot be addressed directly and requires an attributePath:
+//
+//	plancheck.ExpectUnknownOutputValue(plancheck.OutputValueParams{
+//	    OutputAddress: "string_attribute",
+//	    AttributePath: tfjsonpath.New("rfc3339"),
+//	}),
+type OutputValueParams struct {
+	OutputAddress string
+	AttributePath tfjsonpath.Path
+}
```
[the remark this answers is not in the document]
bendbennett: I've updated to `ValuePath` for now, but let's discuss potentially splitting into separate functions.

plancheck/expect_unknown_output_value.go
bendbennett: I've added the documentation for `OutputValueParams` to the website docs. It now appears in both the website docs and in the Go package docs.

website/docs/plugin/testing/acceptance-tests/plan-checks.mdx
bendbennett: Have split the page into Plan Checks - Overview, Resource, Output, and Custom.

The docs for `ExpectEmptyPlan` and `ExpectNonEmptyPlan` currently appear on the Resource page. Perhaps these should be on a separate page or on the Overview page?

bendbennett:
> More generally, I think we should come up with common language to differentiate between attribute state and attribute output/value. Especially with `expectUnknownValue()` and `expectUnknownOutputValue()` where the language is a bit too similar. Maybe we can describe `expectUnknownValue()` and `expectSensitiveValue()` as "asserting that an attribute is being marked as unknown/sensitive."

I've modified the wording on `ExpectUnknownOutputValue`, and `ExpectNullOutputValue` to emphasise the difference. For instance:

> ExpectUnknownOutputValue returns a plan check that asserts that the specified output has an unknown value.

Do you think this provides enough distinction/clarity?

plancheck/expect_unknown_output_value.go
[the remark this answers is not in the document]
bendbennett: Following out-of-band discussion, the `Expect<Null|Unknown>OutputValue` funcs have been split into:

- `Expect<Null|Unknown>OutputValue`
- `Expect<Null|Unknown>OutputValueAtPath`

website/docs/plugin/testing/acceptance-tests/plan-checks.mdx
bflad: Also ❤️ for thinking ahead to just split up the pages.

bflad (approved):
Overall, this is wonderful. Looks good to me 🚀

plancheck/expect_unknown_output_value_at_path_test.go
```diff
@@ -0,0 +1,367 @@
...
+
+	r.Test(t, r.TestCase{
+		ExternalProviders: map[string]r.ExternalProvider{
+			"time": {
+				Source: "registry.terraform.io/hashicorp/time",
+			},
+		},
+		Steps: []r.TestStep{
+			{
+				Config: `resource "time_static" "one" {}
```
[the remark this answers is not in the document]
bflad: Well, this is fun. We validate things a little _too_ well currently if you fully omit referencing some provider in the test case/step configuration.

```
Test validation error: TestStep 1/2 validation error: Providers must be specified at the TestCase level, or in all TestStep, or in TestStep.ConfigDirectory or TestStep.ConfigFile
```

I'll create a separate issue to see if we can make it possible to create testing with just the built-in `terraform` provider.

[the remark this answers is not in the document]
bflad: Ah ha -- found a workaround based on a [special provider discovery rule](https://developer.hashicorp.com/terraform/language/providers/requirements#built-in-providers):

```go
ExternalProviders: map[string]resource.ExternalProvider{
	"terraform": {Source: "terraform.io/builtin/terraform"},
},
```

That works as expected although it calls `terraform init` unnecessarily in this very specific case -- I'll look to see if it makes sense to potentially have the `init` calling logic skip if `terraform.io/builtin/terraform` is the only external provider and source.

website/docs/plugin/testing/acceptance-tests/plan-checks.mdx
bendbennett: Have added a `General Plan Checks` section to the Overview page to include plan checks for `ExpectEmptyPlan`/`ExpectNonEmptyPlan`.

plancheck/expect_unknown_output_value_at_path_test.go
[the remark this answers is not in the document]
bendbennett: Nice! I've replaced the usage of the time provider with the built-in terraform_data resource in all of the `ExpectUnknownOutputValue`, and `ExpectUnknownOutputValueAtPath` plan check tests.

website/docs/plugin/testing/acceptance-tests/plan-checks/index.mdx
```diff
@@ -0,0 +1,22 @@
...
+These `terraform plan` operations results in a [plan file](/terraform/cli/commands/plan#out-filename) and can be represented by this [JSON format](/terraform/internals/json-format#plan-representation).
+
+A **plan check** is a test assertion that inspects the plan file at a specific phase during the current testing mode. Multiple plan checks can be run at each defined phase, all assertion errors returned are aggregated, reported as a test failure, and all test cleanup logic is executed.
+
+- Available plan phases for **Lifecycle (config)** mode are defined in the [`TestStep.ConfigPlanChecks`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep) struct
+- Available plan phases for **Refresh** mode are defined in the [`TestStep.RefreshPlanChecks`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep) struct
+- **Import** mode currently does not run any plan operations, and therefore does not support plan checks.
+
+There are built-in [resource plan checks](/terraform/plugin/testing/acceptance-tests/plan-checks/resource), and [output plan checks](/terraform/plugin/testing/acceptance-tests/plan-checks/output).
```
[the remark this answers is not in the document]
bendbennett: I've updated the Overview and Resource pages to explicitly mention "managed resources, and data sources". I went with "data sources" rather than "data resources" as it seems likely that the former might be more familiar terminology given the documentation on the terraform language for [Data Sources](https://developer.hashicorp.com/terraform/language/data-sources).

Issue #222, opened 2023-11-22
plancheck: ExpectEmptyPlan Does Not Account for Output Changes


bflad:
### terraform-plugin-testing version

```
v1.5.1
```

### Relevant provider source code

```go
resource.UnitTest(t, resource.TestCase{
  Steps: []resource.TestStep{
    {
      Config: `output "test" { value = "one" }`,
    },
    {
      Config: `output "test" { value = "two" }`,
      ConfigPlanChecks: resource.ConfigPlanChecks{
        PreApply: []plancheck.PlanCheck{
          plancheck.ExpectEmptyPlan(),
        },
      },
    },
  },
})
```

### Expected Behavior

Given the plan check naming and that in the human-readable plan output that the `output` change would show, that this assertion would fail the test.

### Actual Behavior

Assertion passes and test passes.

### Steps to Reproduce

`go test`

### References

- #221 (fixing the behavior in this report would obviate some need for a separate plan check, although it would still have its own purpose)

bflad:
Given the similar resource changes only checking of `ExpectNonEmptyPlan`, I would also presume this applies to that plan check as well in the inverse way.

bflad:
So this is actually very nuanced. The testing logic currently always calls a `refresh` command before `plan`. That refresh will update the output value so the plan views it as a no-op action. The explicit refresh does not need to explicitly be there as it can be done on the plan. Further down the rabbit hole we go!

Issue #194, opened 2023-09-29
Tests are executed 5 times each time


gerbil:
### terraform-plugin-testing version
```
v1.5.1
```

### Relevant provider source code

```go

package opswatProvider

import (
	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
	"regexp"
	"testing"
)

func TestAccFileSyncDataSource(t *testing.T) {
	resource.Test(t, resource.TestCase{
		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
		Steps: []resource.TestStep{
			// Read testing
			{
				Config: providerConfig + configFileSyncDataSource,
				Check: resource.ComposeTestCheckFunc(
					// Verify value returned
					resource.TestMatchResourceAttr("data.opswat_file_sync.test", "timeout", regexp.MustCompile("[0-9]+")),
				),
			},
		},
	})
}

const configFileSyncDataSource = `
data "opswat_file_sync" "test" {}
output "opswat_file_sync" {
  value = data.opswat_file_sync.test
}
`

```

### Terraform Configuration Files
```tf
data "XXX" "test" {}
output XXX" {
  value = data.XXX.test
}
```

### Expected Behavior
Test should be executed only once

### Actual Behavior
Test was executed 5 times (based on fmt.Println in GetGlobalSync() func)

$ TF_ACC=1 go test ./internal/test/ -v
=== RUN   TestAccFileSyncDataSource
request URL: https://xxx/admin/config/file/sync
request URL: https://xxx/admin/config/file/sync
request URL: https://xxx/admin/config/file/sync
request URL: https://xxx/admin/config/file/sync
request URL: https://xxx/admin/config/file/sync
--- PASS: TestAccFileSyncDataSource (21.23s)
PASS
ok      terraform-provider-opswat/internal/test 23.468s

### Steps to Reproduce
git clone https://github.com/gerbil/terraform-provider-opswat 
$ TF_ACC=1 go test ./internal/test/ -v

gerbil:
Btw applicable only to datasource tests

bflad:
Hi @gerbil 👋 Thank you for submitting this issue.

[...]

PR #223, opened 2023-11-22
helper/resource: Refresh during plan rather than separately and remove state shimming when possible

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-sdk/pull/515
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/194
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/222
Reference: https://developer.hashicorp.com/terraform/cli/commands/refresh

This change replaces the usage of explicit refresh commands by instead calling the next command with the relevant refresh flags. Separating the refresh from plan was previously done with the idea that the testing logic may introduce further assertion hooks based on that explicit refresh. Over time though, the Terraform recommendations for refreshing and planning have solidified around always recommending to use a plan with behavior changing flags as necessary. Even the Terraform CLI implementation of the refresh command itself has been shifted to call a plan with special flags.

This should reduce testing times decently as it means that the testing logic will skip spinning up and tearing down any relevant providers two times per `TestStep` and whatever time would be associated with Terraform CLI loading for those commands. Comparing the testing performed across this Go module:

Prior: `TF_ACC=1 go test -count=1 ./...  111.07s user 59.37s system 233% cpu 1:12.99 total`
Now: `TF_ACC=1 go test -count=1 ./...  90.81s user 48.25s system 229% cpu 1:00.54 total`

The separated refresh also prevented `PreApply` plan checks from catching certain situations because the refreshed state could cause the plan to be missing changes.

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20231122-115402.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20231122-115402.yaml
@@ -1,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'helper/resource: Removed separate refresh commands, which increases testing
+  performance'
+time: 2023-11-22T11:54:02.542726-05:00
+custom:
+  Issue: "223"
--- a/helper/resource/testing_new_config.go
+++ b/helper/resource/testing_new_config.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hashicorp/terraform-exec/tfexec"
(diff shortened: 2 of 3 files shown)


bendbennett (approved):
LGTM 🚀

bflad:
If they have time, I'm soliciting a `hashicorp/aws` acceptance test run with this change before merging to feel even more confident about it.

bflad:
If any bugs are caused by this change, please open a bug report and we will take a look.

PR #224, opened 2023-11-22
ci: Enable acceptance testing

bflad:
This change adds Terraform version matrix testing to the Go code testing workflow and ensures that acceptance testing is enabled, since there are many tests that require the environment variable to run. Unsetting `TF_ACC_TERRAFORM_PATH` is necessary for locally running the tests on other Terraform versions.


bendbennett (approved):
LGTM 🚀

PR #228, opened 2023-11-29
Skip tests when resource unavailable or entire resource is marked unknown

bendbennett:
### Background

Recently, the `terraform-plugin-testing` module was updated to use a matrix of Terraform versions to use when running tests (refer to [ci: Enable acceptance testing](https://github.com/hashicorp/terraform-plugin-testing/pull/224#top)). This has uncovered two issues:

- Tests that use `terraform_data` fail, as the `terraform_data` resource is only available in Terraform `v1.4.0` and later versions (refer to [Enhancements](https://github.com/hashicorp/terraform/blob/v1.4/CHANGELOG.md#140-march-08-2023)).
- Tests that use a path within an output value fail if any attribute value within the resource is unknown, as the entire resource is then marked as unknown in Terraform versions prior to `v1.3.0` (refer to [Upgrade Notes](https://github.com/hashicorp/terraform/blob/v1.3/CHANGELOG.md#130-september-21-2022)).  

#### Using `terraform_data` when resource is not available

```shell
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 12, in resource "terraform_data" "one":
          12: resource "terraform_data" "one" {
        
        The provider terraform.io/builtin/terraform does not support resource type
        "terraform_data".
```

#### Using an output value plan check when entire resource is marked unknown

```shell
        path not found: cannot convert object at MapStep string_attribute to map[string]any
```

The following illustrates the difference between Terraform versions < `v1.3.0`, and  >= `v1.3.0`:

##### Terraform prior to `v.1.3.0`

image

##### Terraform `v1.3.0` or later

image
 
#### Use Terraform version checks for `terraform_data`

```go
		// The terraform_data resource is not available prior to Terraform v1.4.0
		// Reference: https://github.com/hashicorp/terraform/blob/v1.4/CHANGELOG.md#140-march-08-2023
		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
			tfversion.SkipBelow(version.Must(version.NewVersion("1.4.0"))),
		},
```

#### Use Terraform version checks for output value

```go
		// Prior to Terraform v1.3.0 a planned output is marked as fully unknown
		// if any attribute is unknown. The id attribute within the test provider
		// is unknown.
		// Reference: https://github.com/hashicorp/terraform/blob/v1.3/CHANGELOG.md#130-september-21-2022
		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
			tfversion.SkipBelow(version.Must(version.NewVersion("1.3.0"))),
		},
```

### DIFF
--- a/plancheck/expect_null_output_value_at_path_test.go
+++ b/plancheck/expect_null_output_value_at_path_test.go
@@ -7,17 +7,26 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
 	r "github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
 func Test_ExpectNullOutputValueAtPath_StringAttribute_EmptyConfig(t *testing.T) {
 	t.Parallel()
 
 	r.Test(t, r.TestCase{
+		// Prior to Terraform v1.3.0 a planned output is marked as fully unknown
+		// if any attribute is unknown. The id attribute within the test provider
+		// is unknown.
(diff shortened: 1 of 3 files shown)


bflad (approved):
Looks good to me 🚀 Suggest using available `tfversion` package variables to remove go-version import and slightly save some compilation time.

Issue #221, opened 2023-11-21
Consider Plan Check for Output Value Changes


bflad:
### terraform-plugin-testing version

```
v1.5.1
```

### Use cases

There are scenarios where developers may want to use output values for checking plan differences rather than needing a "heavier" configuration setup of a managed resource to store the state and check against that managed resource for plan differences. Checking whole data source contents between provider versions is one such example.

For that example, the configuration looks like:

```terraform
data "examplecloud_thing" "test" {}

resource "terraform_data" "test" {
  input = data.examplecloud_thing.test
}
```

If using the legacy `null_resource` managed resource, there's additional testing configuration necessary such as properly setting up the `TestStep.ExternalProviders` field.

Developers may not be contextually expecting that a managed resource is the correct way to store a value of a provider concept which implicitly does not store state across Terraform runs, instead reaching for an output, which can accomplish the same thing:

```terraform
data "examplecloud_thing" "test" {}

output "test" {
  value = data.examplecloud_thing.test
}
```

When testing this way via `TestStep.PlanOnly` though, output value changes do not cause the test to fail. Instead a plan check must be used, which there is no built-in plan check available today.

### Proposal

At some point, we should likely consider deprecating `PlanOnly` since its a confusing abstraction for developers and instead recommend plan checks. This will be separately proposed though.

This Go module could introduce a new built-in plan check for this, e.g. as a code sketch

```go
// naming debatable of course, this is just a working code sketch
func ExpectOutputNoChanges(output string) plancheck.PlanCheck {
	return expectOutputNoChangesCheck{
		output: output,
	}
}

type expectOutputNoChangesCheck struct {
	output string
}

func (c expectOutputNoChangesCheck) CheckPlan(ctx context.Context, req plancheck.CheckPlanRequest, resp *plancheck.CheckPlanResponse) {
	if len(req.Plan.OutputChanges) == 0 {
		return
	}

	change, ok := req.Plan.OutputChanges[c.output]

	if !ok || change == nil {
		return
	}

	resp.Error = fmt.Errorf("unexpected output %q change", c.output)
}
```

This then plan check is usable for this testing situation:

```go
ConfigPlanChecks: resource.ConfigPlanChecks{
	PreApply: []plancheck.PlanCheck{
		plancheck.ExpectOutputNoChanges("test"),
	}.
}
```

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/875

PR #232, opened 2023-11-29
plancheck: Ensure `ExpectEmptyPlan` and `ExpectNonEmptyPlan` account for output changes in addition to resource changes

### LINKED ISSUE #222: plancheck: ExpectEmptyPlan Does Not Account for Output Changes

bflad:
Closes #222

Previously with new unit tests:

```
--- FAIL: Test_ExpectNonEmptyPlan_OutputChanges_None (0.90s)
    expect_non_empty_plan_test.go:17: Step 2/2 error: Pre-apply plan check(s) failed:
        expected a non-empty plan, but got an empty plan
--- FAIL: Test_ExpectEmptyPlan_OutputChanges_Error (1.00s)
    expect_empty_plan_test.go:41: Step 2/2, expected an error but got none
```

### DIFF
--- a/plancheck/expect_empty_plan_test.go
+++ b/plancheck/expect_empty_plan_test.go
@@ -9,27 +9,22 @@ import (
 
 	r "github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func Test_ExpectEmptyPlan(t *testing.T) {
+func Test_ExpectEmptyPlan_OutputChanges_None(t *testing.T) {
 	t.Parallel()
 
-	r.Test(t, r.TestCase{
+	r.UnitTest(t, r.TestCase{
 		ExternalProviders: map[string]r.ExternalProvider{
-			"random": {
-				Source: "registry.terraform.io/hashicorp/random",
-			},
+			"terraform": {Source: "terraform.io/builtin/terraform"},
 		},
 		Steps: []r.TestStep{
 			{
-				Config: `resource "random_string" "one" {
-					length = 16
-				}`,
+				Config: `output "test" { value = "original" }`,
 			},
 			{
-				Config: `resource "random_string" "one" {
-					length = 16
-				}`,
+				Config: `output "test" { value = "original" }`,
 				ConfigPlanChecks: r.ConfigPlanChecks{
 					PreApply: []plancheck.PlanCheck{
 						plancheck.ExpectEmptyPlan(),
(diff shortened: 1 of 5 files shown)


plancheck/expect_empty_plan_test.go
```diff
@@ -9,74 +9,128 @@ import (
...
-	r.Test(t, r.TestCase{
+	r.UnitTest(t, r.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.SkipBelow(tfversion.Version0_14_0), // outputs before 0.14 always show as created
+		},
+		// Avoid our own validation that requires at least one provider config.
 		ExternalProviders: map[string]r.ExternalProvider{
-			"random": {
-				Source: "registry.terraform.io/hashicorp/random",
+			"terraform": {Source: "terraform.io/builtin/terraform"},
```
austinvalle, line 24: TIL this was a thing lol

austinvalle (approved):
LGTM, just wanted to double check [this](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks#built-in-plan-checks) wording on the docs site: 

> _Asserts the entire plan has no operations for apply._
> _Asserts the entire plan contains at least one operation for apply._

This still sounds correct? If we count output value changes as "an operation for apply"

plancheck/expect_non_empty_plan_test.go
```diff
@@ -9,26 +9,83 @@ import (
...
+						plancheck.ExpectNonEmptyPlan(),
+					},
+				},
+				ExpectError: regexp.MustCompile(`expected a non-empty plan, but got an empty plan`),
+			},
+		},
+	})
+}
+
+func Test_ExpectNonEmptyPlan_ResourceChanges_None(t *testing.T) {
```
bendbennett, line 70: Nit: Should `_None` be used here? Seems like it's indicating that there are no resource changes?

plancheck/expect_non_empty_plan_test.go
```diff
@@ -9,26 +9,83 @@ import (
 
 	r "github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func Test_ExpectNonEmptyPlan(t *testing.T) {
+func Test_ExpectNonEmptyPlan_OutputChanges_None(t *testing.T) {
```
bendbennett, line 15: Nit: Should `_None` be used here? Seems like it's indicating that there are no output changes?

bendbennett (approved):
LGTM 🚀

plancheck/expect_non_empty_plan_test.go
bflad: Good shout -- updated the test naming.

bflad:
@austinvalle great question re: If we count output value changes as "an operation for apply"

I believe we should. Output values are planned as differences and saved into state via apply since they affect consuming modules referencing the value directly or via `terraform_remote_state`. I think this is the beauty of the `plancheck` interface -- if folks need more granular checking, e.g. only error on resource changes, we can implement that with an explicitly named check. I'll update the title and description of #221 to capture that better.

PR #234, opened 2023-11-30
helper/resource: Ensure TestStep.ExpectNonEmptyPlan accounts for output changes

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/222

This is a followup to similar `plancheck` implementation updates that ensure output changes are checked in the plan in addition to resource changes. The intention of this functionality is to catch any plan differences and has been documented in this manner for a long while.

Since `TestStep.ExpectNonEmptyPlan` pre-dates this Go module, for extra compatibility for developers migrating, output checking is only enabled for Terraform 0.14 and later since prior versions will always show changes when outputs are present in the configuration. Ideally `TestStep.ExpectNonEmptyPlan` will be deprecated in preference of the newer plan checks since its abstraction is fairly ambiguous to when its being invoked, but this deprecation may come by nature of larger changes for a next major version.

New unit test failure prior to implementation update:

```
--- FAIL: Test_ExpectNonEmptyPlan_PostRefresh_OutputChanges (0.42s)
    /Users/bflad/src/github.com/hashicorp/terraform-plugin-testing/helper/resource/testing_new_config_test.go:122: Step 1/1 error: Expected a non-empty plan, but got an empty plan
```

### DIFF
--- a/helper/resource/testing_new_refresh_state.go
+++ b/helper/resource/testing_new_refresh_state.go
@@ -88,7 +88,7 @@ func testStepNewRefreshState(ctx context.Context, t testing.T, wd *plugintest.Wo
 		}
 	}
 
-	if !planIsEmpty(plan) && !step.ExpectNonEmptyPlan {
+	if !planIsEmpty(plan, wd.GetHelper().TerraformVersion()) && !step.ExpectNonEmptyPlan {
 		var stdout string
 		err = runProviderCommand(ctx, t, func() error {
 			var err error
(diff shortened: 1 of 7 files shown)


bflad:
To be honest, I would also be fine only adding the missing testing here and leaving the existing `TestStep.ExpectNonEmptyPlan` implementation. This change is being suggested for consistency between the two ways of checking plans.

helper/resource/testing_new_refresh_state.go
austinvalle, line 1: Is it possible to add a test for this 😃

austinvalle (approved):
LGTM! I like updating both implementations to be consistent. I'm not sure how often output values are used in provider tests, but it seems a very reasonable bug fix to make now.

Just had two notes in the PR :shipit:

bendbennett (approved):
LGTM just a couple of minor comments.

PR #238, opened 2023-12-05
helper/resource: Clarify configuration step error messages

bflad:
When using `TestStep.PlanOnly`, the error messaging errantly mentions:

```
After applying this test step, the plan was not empty.
```

Which is inaccurate and confusing for debugging. This also adjusts error messages to include refresh vs non-refresh wording consistently. The error messages have no compatibility promises and it would be considered an errant provider developer implementation to rely on them, rather than properly fixing the test assertions (such as using `TestStep.ExpectNonEmptyPlan` instead of `TestCase.ErrorCheck` or `TestStep.ExpectError` against the error messaging without compatibility promises).


bendbennett (approved):
LGTM 🚀 

Wondering whether we might want to consider [modifying the documentation](https://github.com/hashicorp/terraform-plugin-testing/issues/240) to illustrate which specific terraform commands are being run when `config`, `import`, or `refresh` modes are being used?

bflad:
Looks like you already created https://github.com/hashicorp/terraform-plugin-testing/pull/238 so we can discuss that proposal over there.

PR #251, opened 2023-12-19
website: Additional terraform.InstanceState and resource.Retry clarifications in migration guide

bflad:
These items came up during an official provider migration which:

- Implemented terraform-plugin-sdk based resource with state migrations and covering unit testing that needed to remain.
- Had not migrated `resource.Retry` to `resource.RetryContext` yet, so needed the extra entry


austinvalle (approved):
🚀

Issue #256, opened 2023-12-22
`Failed to marshal state to json` error when upgrading from 1.5.1 to 1.6.0


AgustinBettati:
### terraform-plugin-testing version
```
1.6.0
```

### Relevant provider source code

Migration test that uses an old version `1.11.0` of our provider and then runs a plan operation using latest build.

```go
func TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_maintainsBackwardCompatibility(t *testing.T) {
	var (
		...
	)

	resource.ParallelTest(t, resource.TestCase{
		PreCheck:     func() { acc.PreCheckBasic(t) },
		CheckDestroy: acc.CheckClusterDestroy,
		Steps: []resource.TestStep{
			{
				ExternalProviders: map[string]resource.ExternalProvider{
					"mongodbatlas": {
						VersionConstraint: "1.11.0",
						Source:            "mongodb/mongodbatlas",
					},
				},
				Config:            config,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckMongoDBAtlasClusterExists(resourceName, &cluster),
					testAccCheckMongoDBAtlasClusterAttributes(&cluster, name),
					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
					resource.TestCheckResourceAttr(resourceName, "name", name),
				),
			},
			{
				ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
				Config:                   config,
				ConfigPlanChecks: resource.ConfigPlanChecks{
					PostApplyPreRefresh: []plancheck.PlanCheck{
						acc.DebugPlan(),
					},
				},
				PlanOnly: true,
			},
		},
	})
}

```

Relevant attribute definition from version 1.11.0, which was removed in version 1.12.0

```
.....,
"bi_connector": {
        Type:          schema.TypeMap,
        Optional:      true,
        Deprecated:    "use bi_connector_config instead",
        ConflictsWith: []string{"bi_connector_config"},
        Elem: &schema.Schema{
	        Type: schema.TypeString,
        },
},
```

### Terraform Configuration Files

Note that `bi_connector` attribute is not defined.
```tf
resource "mongodbatlas_cluster" "test" {
	project_id                   = mongodbatlas_project.cluster_project.id
	name                         = %[3]q
	disk_size_gb                 = 100
        cluster_type = "REPLICASET"
        replication_specs {
            num_shards = 1
            regions_config {
[...]

Issue #237, opened 2023-12-05
Error with state upgrades after `terraform-plugin-testing@v1.6.0` update


ewbankkit:
### terraform-plugin-testing version

```
v1.6.0
```

##### Terraform Schema

```go
		resourceV0 := &schema.Resource{Schema: map[string]*schema.Schema{}}

		SchemaVersion: 1,
		StateUpgraders: []schema.StateUpgrader{
			{
				Type: resourceV0.CoreConfigSchema().ImpliedType(),
				Upgrade: func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
					if v, ok := rawState["enable_default_standards"]; !ok || v == nil {
						rawState["enable_default_standards"] = "true"
					}

					return rawState, nil
				},
				Version: 0,
			},
		},
		Schema: map[string]*schema.Schema{
			"enable_default_standards": {
				Type:     schema.TypeBool,
				Optional: true,
				ForceNew: true,
				Default:  true,
			},
		},

```

##### Acceptance Test Case

```go
func testAccAccount_migrateV0(t *testing.T) {
	ctx := acctest.Context(t)
	resourceName := "aws_securityhub_account.test"

	resource.Test(t, resource.TestCase{
		PreCheck:     func() { acctest.PreCheck(ctx, t) },
		ErrorCheck:   acctest.ErrorCheck(t, names.SecurityHubEndpointID),
		CheckDestroy: testAccCheckAccountDestroy(ctx),
		Steps: []resource.TestStep{
			{
				ExternalProviders: map[string]resource.ExternalProvider{
					"aws": {
						Source:            "hashicorp/aws",
						VersionConstraint: "4.59.0",
					},
				},
				Config: testAccAccountConfig_basic,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckAccountExists(ctx, resourceName),
					resource.TestCheckNoResourceAttr(resourceName, "enable_default_standards"),
				),
			},
			{
				ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories,
				Config:                   testAccAccountConfig_basic,
				PlanOnly:                 true,
			},
		},
	})
}
```

### Terraform Configuration Files

```tf
resource "aws_securityhub_account" "test" {}
```

### Expected Behavior

With **v1.5.1**

```console
% make testacc TESTARGS='-run=TestAccSecurityHub_serial/^Account$$/MigrateV0' PKG=securityhub
[...]

PR #257, opened 2023-12-26
Additional CHANGELOG note for v1.6.0

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/pull/223
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/237
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/256

Post-release update to the v1.6.0 CHANGELOG to capture now-known situations that would affect existing provider testing. This update ensures developers upgrading to v1.6.0 will receive the latest guidance and the guidance will remain for this version after future releases. If approved and merged, the v1.6.0 GitHub Release description will also be updated with the same additional content.


austinvalle (approved):
lgtm 🚀

PR #258, opened 2023-12-26
deps: Remove direct dependency on github.com/mitchellh/copystructure

bflad:
Reference: https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc

Its usage was only in esoteric functionality that is unintentionally exported and already deprecated. The `maps`/`slices` Go standard library packages are only Go 1.21+ so could not leverage that immediately since this Go module is 1.20+ until 1.22 is released, but the structures are small enough that its not too much lift to implement the manual copying code.

Remaining indirect dependency is via sdk/v2 (which that dependency will go away next major version anyways):

```
# github.com/mitchellh/copystructure
github.com/hashicorp/terraform-plugin-testing/helper/resource
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema
github.com/mitchellh/copystructure
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/hashicorp/terraform-plugin-go v0.20.0
 	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.31.0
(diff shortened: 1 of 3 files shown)


austinvalle (approved):
🚀

PR #259, opened 2023-12-26
deps: Remove direct dependency on github.com/mitchellh/reflectwalk

bflad:
Reference: https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc

Its usage was only in esoteric functionality that is unintentionally exported and already deprecated.

Remaining indirect dependency is via sdk/v2 (which that dependency will go away next major version):

```
# github.com/mitchellh/reflectwalk
github.com/hashicorp/terraform-plugin-testing/helper/resource
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema
github.com/hashicorp/terraform-plugin-sdk/v2/terraform
github.com/mitchellh/reflectwalk
```


austinvalle (approved):
🧹 🧹

PR #260, opened 2023-12-26
deps: Remove golang.org/x/exp dependency

bflad:
Reference: https://github.com/golang/go/issues/61914

It was only used for the `constraints` package, the type definitions are relatively trivial and static, and the Go language maintainers are not sure if/where these will live in the standard library, which implies (at least to me) they could also disappear from x/exp.


austinvalle (approved):
:shipit:

PR #261, opened 2023-12-28
website: Mention PlanOnly to ConfigPlanChecks in migration guide

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/237
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/256

Technically, this testing change is always safe to make, but this calls out the currently known situations it will cause errors to not over-burden developers during their migration.


bendbennett (approved):
LGTM 🚀

Issue #264, opened 2024-01-08
Modify detection of provider blocks in configuration


bendbennett:
### terraform-plugin-testing version
```
v1.6.0
```

### Relevant provider source code

N/A

### Terraform Configuration Files

The following are all allowed `provider` block declarations from the perspective of `terraform validate`.

```tf
provider"random"{}
```
```tf
provider "random"{}
```
```tf
provider random{}
```

### Expected Behavior
Using the `provider` block configurations supplied should result in successfully executing tests.

### Actual Behavior
The following error is raised:

```shell
        Error: Duplicate provider configuration
        
          on terraform_plugin_test.tf line 13:
          13: provider"random"{}
        
        A default (non-aliased) provider configuration for "random" was already given
        at terraform_plugin_test.tf:11,1-18. If multiple configurations are required,
        set the "alias" argument for alternative configurations.
    testing_new.go:80: Error retrieving state, there may be dangling resources: exit status 1
```

### Steps to Reproduce
Run an acceptance test with a `provider` block configuration that uses the quoting and /or spacing described in `Terraform Configuration Files`.

bflad:
I wonder if we should submit something upstream for `provider"random"{}` -- its probably fine for us to support the lack of whitespace after the provider block label, but the label-without-prior-whitespace feels very 😬 to me personally.

bendbennett:
There's actually no usage of `provider"random"{}` in the `random` provider. @austinvalle uncovered that these various forms of provider block declaration are all legitimate terraform configuration, so I thought it might be best to have the detection of these forms of declaration be taken into account in case they caused the errors described.

bflad:
I have no issues with implementing the unquoted and whitespace missing after label cases (e.g. `provider "LABEL"{`), but I'm asking if we should submit an issue to https://github.com/hashicorp/terraform/issues to consider having `terraform validate` raise an error if it encounters `provider"LABEL"{` without the whitespace before the label. It may be valid in HCL for whatever reason, but seems like an awkward Terraform configuration style to support.

austinvalle:
drive-by - 🚗 -  `terraform fmt` will add those whitespaces in (which our addition of config files and directories will make it easier to take advantage of 🥳 )

![image](https://github.com/hashicorp/terraform-plugin-testing/assets/8650838/b7550788-b0ef-42ed-b20b-4e8ef01aeadc)

[...]

PR #265, opened 2024-01-08
Modify regex to detect other legitimate forms of provider block declaration

bendbennett:
Closes: #264

### DIFF
--- a/internal/teststep/config.go
+++ b/internal/teststep/config.go
@@ -21,7 +21,7 @@ const (
 )
 
 var (
-	providerConfigBlockRegex  = regexp.MustCompile(`provider "?[a-zA-Z0-9_-]+"? {`)
+	providerConfigBlockRegex  = regexp.MustCompile(`provider(\s*"[a-zA-Z0-9_-]+"\s*|\s+[a-zA-Z0-9_-]+\s*){`)
 	terraformConfigBlockRegex = regexp.MustCompile(`terraform {`)
 )
 
--- a/internal/teststep/config.go
+++ b/internal/teststep/config.go
@@ -21,6 +21,12 @@ const (
 )
 
 var (
+	// Expected to match:
+	//     provider "example" {
+	//     provider "example"{
+	//     provider example {
+	//     provider example{
+	//     provider"example"{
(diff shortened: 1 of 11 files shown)


internal/teststep/config.go
```diff
@@ -21,7 +21,7 @@ const (
 )
 
 var (
-	providerConfigBlockRegex  = regexp.MustCompile(`provider "?[a-zA-Z0-9_-]+"? {`)
+	providerConfigBlockRegex  = regexp.MustCompile(`provider(\s*"[a-zA-Z0-9_-]+"\s*|\s+[a-zA-Z0-9_-]+\s*){`)
```
bflad, line 24: Nit: Might be good to document (or consider unit testing the regular expression itself, although your other testing seems to cover it well already) the various cases this is intended to catch since its now more complicated:

```suggestion
	// Expected to match:
	//     provider "example" {
	//     provider "example"{
	//     provider example {
	//     provider example{
	//     provider"example"{
	providerConfigBlockRegex  = regexp.MustCompile(`provider(\s*"[a-zA-Z0-9_-]+"\s*|\s+[a-zA-Z0-9_-]+\s*){`)
```

internal/teststep/config.go
```diff
@@ -21,7 +21,7 @@ const (
...
 	terraformConfigBlockRegex = regexp.MustCompile(`terraform {`)
```
bflad, line 25: FYI: I presume the same whitespace problem can occur with `terraform{` as well, although `terraform validate` may not raise an error and Terraform may instead merge the `terraform` configuration block... configurations.

bflad (approved):
Seems reasonable to me 🚀

internal/teststep/config.go
bendbennett: Good call. I can verify that having duplicate `terraform` blocks can lead to issues. For example:

```shell
        Error: Duplicate required providers configuration
        
          on terraform_plugin_test.tf line 13, in terraform:
          13:   required_providers {
        
        A module may have only one required providers configuration. The required
        providers were previously configured at terraform_plugin_test.tf:3,3-21.
    testing_new.go:80: Error retrieving state, there may be dangling resources: exit status 1
```

Have modified the regex for detection of `terraform` block, and updated the tests.

austinvalle (approved):
awesome! 🚀

Issue #266, opened 2024-01-08
Consider Adding State Check Equivalents to Plan Checks


bendbennett:
### terraform-plugin-testing version

```
v.1.6.0
```

### Use cases

The [plancheck](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck) package within `terraform-plugin-testing` has plan checks which use [tfjson.Plan](https://pkg.go.dev/github.com/hashicorp/terraform-json#Plan). This provides structured data which is easily traversed and interrogated for the purposes of, for instance, verifying that a specific resource attribute, or output value has a known type and value.

There is an equivalent [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State) which is available, and would provide an similarly straightforward way to traverse and interrogate the state for the purposes of verifying, for instance, that a specific resource attribute, or output value has a known type and value.

### Attempted solutions

Currently, values in state can be checked by configuring the [TestStep.Check](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep) field and using either predefined [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) functions (e.g., [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckNoResourceAttr)) or by writing custom [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) functions. Checks that use [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) functions are constrained by the flatmap syntax that is required to access state values, and by the fact that all values are represented as strings.

### Proposal

The proposal is to:

- Add, where relevant, state checks that are equivalent to the previously implemented plan checks.
- ~~Add predefined/built-in state checks that provide equivalent functionality to that provided by the predefined [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) functions (e.g., [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckNoResourceAttr)).~~

### Further Considerations

[...]

PR #276, opened 2024-01-16
Adding experimental status and example usage for `ExpectKnownValue`, `ExpectKnownOutputValue`, and `ExpectKnownOutputValueAtPath` plan and state checks

bendbennett:
The original intent of this PR has been altered. Rather than deprecating the pre-existing built-in [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) implementations, examples of how to use the newer `ExpectKnownValue`, `ExpectKnownOutputValue`, and `ExpectKnownOutputValueAtPath` plan and state checks, with the `knownvalue` package have been added. 

References: #266 

This PR adds predefined/built-in state checks that provide equivalent functionality to that provided by the predefined [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) functions (e.g., [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckNoResourceAttr)).

The following implementations have been included:

| TestCheckFunc  | State Check |
| ------------- | ------------- |
| [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckNoResourceAttr)  | `ExpectNoValueExists`  |
| [TestCheckOutput](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckOutput)  | `ExpectKnownOutputValue`, `ExpectKnownOutputValueAtPath`  |
| [TestCheckResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttr)  | `ExpectKnownValue`  |
| [TestCheckResourceAttrPair](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttrPair)  | `ExpectMatchingValues`  |
| [TestCheckResourceAttrPtr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttrPtr)  | Can just use a pointer and deference - see [example](https://github.com/hashicorp/terraform-plugin-testing/pull/276/files#diff-d2a964d09322d8e5b113d7e2acf365f34b87fc572bf0489ed25fb6008d2e2431R103)  |
| [TestCheckResourceAttrSet](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttrSet)  | `ExpectValueExists`  |
| [TestCheckResourceAttrWith](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttrWith)  | `Custom` known value type check can be used  |
| [TestCheckTypeSetElemAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckTypeSetElemAttr)  | `SetExact` and `SetPartial` known value type checks can be used  |
| [TestCheckTypeSetElemAttrPair](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckTypeSetElemAttrPair)  | `ExpectContains`  |
| [TestCheckTypeSetElemNestedAttrs](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckTypeSetElemNestedAttrs)  | `SetExact` and `SetPartial` known value type checks can be used   |
| [TestMatchOutput](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestMatchOutput)  | `ExpectKnownOutputValue`, and `ExpectKnownOutputValueAtPath` with`Custom` known value type check can be used   |
| [TestMatchResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestMatchResourceAttr)  | `ExpectKnownValue` with `Custom` known value type check can be used  |
| [TestMatchTypeSetElemNestedAttrs](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestMatchTypeSetElemNestedAttrs)  | `Set` or `List` known value type checks with nested `Custom` known value type checks can be used  |


website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
```diff
@@ -90,13 +110,111 @@ func TestExpectKnownOutputValue_CheckPlan_Bool(t *testing.T) {
 }
 ```
 
-## Example using `plancheck.ExpectUnknownOutputValue`
+## `ExpectNullOutputValue` Plan Check
```
austinvalle, line 113: Do we want to remove this documentation since we've recently decided to deprecate this? Or at least a note suggesting what to use instead.

website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
```diff
@@ -90,13 +110,111 @@ func TestExpectKnownOutputValue_CheckPlan_Bool(t *testing.T) {
...
+						plancheck.ExpectNullOutputValue("string_attribute"),
+					},
+				},
+			},
+		},
+	})
+}
+```
+
+## `ExpectNullOutputValueAtPath` Plan Check
```
austinvalle, line 153: Do we want to remove this documentation since we've recently decided to deprecate this? Or at least a note suggesting what to use instead.

helper/resource/testing.go
```diff
@@ -1040,6 +1058,49 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
...
+//						),
+//					},
+//				},
+//			},
+//		})
+//	}
+//
+// [ExpectValueExists]: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/statecheck#ExpectValueExists
+// [StateCheck]: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/statecheck#StateCheck
+// [statecheck]: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/statecheck
```
austinvalle, line 1103: ❤️

helper/resource/testing.go
```diff
@@ -1222,6 +1328,84 @@ func testCheckResourceAttr(is *terraform.InstanceState, name string, key string,
...
 //
 // When this function returns an error, TestCheckResourceAttrWith will fail the check.
+//
+// Deprecated: State checks have been superseded by ConfigStateChecks.
+// Use the built-int statecheck.ExpectKnownValue state check in combination
+// with a custom knownvalue.Check instead.
+// CheckResourceAttrWithFunc will be removed in the next major version.
+//
+// The following is an example of using statecheck.ExpectKnownValue in combination
+// with a custom knownvalue.Check to replicate the behaviour of TestCheckResourceAttrWith.
```
austinvalle, line 1338: I like the code snippet that references how to use the new `statecheck` package, but for showing an implementation of a custom knownvalue.... do we want to just link the website page for that?

Same for the other custom known value references in deprecation notices.

statecheck/expect_contains.go
```diff
@@ -0,0 +1,154 @@
...
+		resp.Error = fmt.Errorf("state is nil")
+	}
+
+	if req.State.Values == nil {
+		resp.Error = fmt.Errorf("state does not contain any state values")
+	}
+
+	if req.State.Values.RootModule == nil {
+		resp.Error = fmt.Errorf("state does not contain a root module")
+	}
```
austinvalle, line 40: Returns after setting an error response 🙂 

(same applies to the other state checks in this PR)

statecheck/expect_contains.go
```diff
@@ -0,0 +1,154 @@
...
+type expectContains struct {
+	resourceAddressOne string
+	attributePathOne   tfjsonpath.Path
+	resourceAddressTwo string
+	attributePathTwo   tfjsonpath.Path
+}
+
+// CheckState implements the state check logic.
+func (e expectContains) CheckState(ctx context.Context, req CheckStateRequest, resp *CheckStateResponse) {
+	var rcOne, rcTwo *tfjson.StateResource
```
austinvalle, line 28: super nit: `rc` -> `r`

(same applies to the other state checks in this PR)

statecheck/expect_contains.go
```diff
@@ -0,0 +1,154 @@
...
+	if req.State.Values == nil {
+		resp.Error = fmt.Errorf("state does not contain any state values")
+	}
+
+	if req.State.Values.RootModule == nil {
+		resp.Error = fmt.Errorf("state does not contain a root module")
+	}
+
+	for _, resourceChange := range req.State.Values.RootModule.Resources {
+		if e.resourceAddressOne == resourceChange.Address {
```
austinvalle, line 43: super nit: `resourceChange` -> `resource`

(same applies to the other state checks in this PR)

statecheck/expect_contains.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		resp.Error = err
+
+		return
+	}
+
+	resultOneCollection, ok := resultOne.([]any)
+
+	if !ok {
+		resp.Error = fmt.Errorf("%s.%s is not a set", e.resourceAddressOne, e.attributePathOne.String())
```
austinvalle, line 89: Is this state check only intended to be used with sets? Could change the error messaging to `"%s.%s is not a collection"`

austinvalle (approved):
Overall LGTM

I'm not sure how I feel about the larger code snippets in Go doc strings. I think for deprecation messages it's nice because we show the alternative right next to the lint itself.

Maybe for some of the examples that are not deprecation notes we could utilize the `example` files where Go will actually compile them (and we will get build errors if the syntax is incorrect). We did this for the recent custom type implementations of `jsontypes`:
- https://go.dev/blog/examples
- Example source: https://github.com/hashicorp/terraform-plugin-framework-jsontypes/blob/main/jsontypes/normalized_value_example_test.go
- Example on the Go doc site: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes#example-Normalized.Unmarshal

website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
bendbennett: Have added a note for now.

website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
bendbennett: Have added a note for now.

bendbennett:
> Overall LGTM
> 
> I'm not sure how I feel about the larger code snippets in Go doc strings. I think for deprecation messages it's nice because we show the alternative right next to the lint itself.
> 
> Maybe for some of the examples that are not deprecation notes we could utilize the `example` files where Go will actually compile them (and we will get build errors if the syntax is incorrect). We did this for the recent custom type implementations of `jsontypes`:
> 
> * https://go.dev/blog/examples
> * Example source: https://github.com/hashicorp/terraform-plugin-framework-jsontypes/blob/main/jsontypes/normalized_value_example_test.go
> * Example on the Go doc site: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes#example-Normalized.Unmarshal

I agree that using Go examples would be a better solution.

I've added examples such as the following:

```go
package statecheck_test

import (
	"testing"

	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
	"github.com/hashicorp/terraform-plugin-testing/statecheck"
	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
)

func ExampleExpectKnownValue() {
	t := &testing.T{}
	t.Parallel()

		// Provider definition omitted.
		Steps: []resource.TestStep{
			{
				Config: `resource "test_resource" "one" {
		          bool_attribute = true
		        }
		        `,
					statecheck.ExpectKnownValue(
						"test_resource.one",
						tfjsonpath.New("bool_attribute"),
						knownvalue.Bool(true),
					),
				},
			},
		},
	})
}
```

It looks a little strange to define `t := &testing.T{}` inline, but this is required for the code to compile, and, as you know, the signature for `func Example<whatever>()` does not allow any arguments. 

Thoughts?

statecheck/expect_matching_values.go
```diff
@@ -0,0 +1,108 @@
...
+// ExpectMatchingValues returns a state check that asserts that the specified attributes at the given resources
+// have a matching value.
+func ExpectMatchingValues(resourceAddressOne string, attributePathOne tfjsonpath.Path, resourceAddressTwo string, attributePathTwo tfjsonpath.Path) StateCheck {
+	return expectMatchingValues{
+		resourceAddressOne: resourceAddressOne,
+		attributePathOne:   attributePathOne,
+		resourceAddressTwo: resourceAddressTwo,
+		attributePathTwo:   attributePathTwo,
+	}
+}
```
bflad, line 108: Similar naming concerns here as I'm a little uneasy about the mixed terminology here:

* Previously "match" meant string regular expression testing
* Previously this 1:1 checking was called "pair"
* Currently with knownvalue checks, they are called "exact"

I'd also consider this along with my prior comment about having an interface/generic check for these sorts of checks, because maybe having something like that would be:

```go
statecheck.ExpectKnownValuePair(
  "examplecloud_thing.test",
  tfjsonpath.New("some_computed_attr"),
  "examplecloud_widget.test",
  tfjsonpath.New("some_other_attr"),
  knownvalue.StringExactPair(),
)
```

And we could still introduce a "wrapper" check if we really wanted.

statecheck/expect_value_exists.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ExpectValueExists returns a state check that asserts that the specified
+// attribute at the given resource exists.
+func ExpectValueExists(resourceAddress string, attributePath tfjsonpath.Path) StateCheck {
+	return expectValueExists{
+		resourceAddress: resourceAddress,
+		attributePath:   attributePath,
+	}
+}
```
bflad, line 75: Open question: Would this be better served with a "not null"/"any value" known value check? I think we can avoid this as its own top-level check so developers don't need to choose their preferred style.

bflad (commented):
Great work so far! Left some upfront considerations we can chat about. 😄

austinvalle:
> It looks a little strange to define `t := &testing.T{}` inline, but this is required for the code to compile, and, as you know, the signature for `func Example<whatever>()` does not allow any arguments.
> 
> Thoughts?

Yeah sometimes it can be a little odd, I'd say if you think it looks weird, just plop a comment on the line explaining it. I think for most cases it'll be understood by the developer.

Example: https://github.com/hashicorp/terraform-plugin-framework-jsontypes/blob/e3e2f72c1e29f0fb91287bdee0b91e73c47ea67c/jsontypes/normalized_value_example_test.go#L25-L29

bflad:
Given we are creating newer/more robust testing documentation (whether in the website or Go package documentation), there's some nuance we may not have called out as much as we could in the past. In this example:

> 			{
>				Config: `resource "test_resource" "one" {
>		          bool_attribute = true
>		        }
>		        `,
>				ConfigStateChecks: []statecheck.StateCheck{
>					statecheck.ExpectKnownValue(
>						"test_resource.one",
>						tfjsonpath.New("bool_attribute"),
>						knownvalue.Bool(true),
>					),
>				},

I think we want to ensure that developers do not feel _required_ do this sort of configuration equals state checking. Especially when developing framework providers, Terraform will error if the state value differs from configuration. The reason the testing logic includes post-apply planning is intended to fail tests for legacy SDK providers for this same problem 😄. In other words, we should generally be encouraging developers (and code examples are a great place for that) to only be state checking _computed_ values. It is not necessarily wrong to check configured values, but its essentially extra work with little to no benefit to the developer.

It may be a little awkward to capture this in a non-confusing way, so suggestions/ideas are very much welcome, but I might suggest something like the following:

```go
{
	// Example resource containing a computed boolean attribute named "computed_attribute"
	Config: `resource "examplecloud_thing" "test" {}`,
		statecheck.ExpectKnownValue(
			"examplecloud_thing.test",
			tfjsonpath.New("computed_attribute"),
			knownvalue.Bool(true),
		),
	},
},
```

Excellent work on all this documentation, so as to surface this sort of thing better!

statecheck/expect_value_exists.go
bendbennett: I think it's likely we'll all agree on using "not null"/"any value" known value check, so I'm implemented that and removed the top-level `ExpectValueExists` check. Happy to refactor/revert if we change our minds.

statecheck/expect_contains.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// Resource State Check
+var _ StateCheck = expectContains{}
+
+type expectContains struct {
+	resourceAddressOne string
+	attributePathOne   tfjsonpath.Path
+	resourceAddressTwo string
+	attributePathTwo   tfjsonpath.Path
```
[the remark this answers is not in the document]
bendbennett: `SetContains` (or whatever comes out of the bikeshedding) could look something like the following:

```go
package knownvalue

import (
	"fmt"
	"reflect"
)

var _ CheckPair = setContains{}

type setContains struct{}

// CheckValue determines whether the passed value is of type []any, and
// contains a matching slice entry.
func (v setContains) CheckValue(a, b any) error {
	collection, ok := a.([]any)

	if !ok {
		return fmt.Errorf("expected []any value for SetContains check, got: %T", a)
	}

	for _, i := range collection {
		if reflect.DeepEqual(i, b) {
			return nil
		}
	}

	return fmt.Errorf("value of %v is not found in value of %v", b, a)
}

// SetContains returns a Check for asserting whether a collection specified
// by the first value passed into Check contains a value specified by the
// second value passed into Check. 
func SetContains() setContains {
	return setContains{}
}
```

statecheck/expect_matching_values.go
bendbennett: I'm in agreement. Think we can drop both `ExpectContain` and `ExpectMatchingValues` in favour of using `ExpectKnownValuePair`.

In terms of exact value matching, I was wondering whether we could forgo type specific value matching and use something like the following:

```go
package knownvalue

import (
	"fmt"
	"reflect"
)

var _ CheckPair = exactPair{}

type exactPair struct{}

// CheckValue determines whether the passed values are equal.
func (v exactPair) CheckValue(a, b any) error {
	if !reflect.DeepEqual(a, b) {
		return fmt.Errorf("values are not equal: %v != %v", a, b)
	}

	return nil
}

// ExactPair returns a Check for asserting whether two values are equal.
func ExactPair() exactPair {
	return exactPair{}
}
```

statecheck/expect_contains.go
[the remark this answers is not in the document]
bendbennett: As discussed out-of-band, have deprecated the `TestCheckResourceAttrPair`, and `TestCheckTypeSetElemAttrPair` functions without replacement and linked to #282 in the docs for now.

statecheck/expect_matching_values.go
bendbennett: As discussed out-of-band, have deprecated the TestCheckResourceAttrPair, and TestCheckTypeSetElemAttrPair functions without replacement and linked to https://github.com/hashicorp/terraform-plugin-testing/issues/282 in the docs for now.

helper/resource/testing.go
```diff
@@ -1612,6 +1921,88 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
...
+//		resource.Test(t, resource.TestCase{
+//			// Provider definition omitted.
+//			Steps: []resource.TestStep{
+//				{
+//					// Example resource containing a computed boolean attribute named "computed_attribute"
+//					Config: `resource "test_resource" "one" {}
+//
+//			        output computed_output {
+//			          value = test_resource.one.computed_attribute
+//			        }`,
```
bflad, line 1955: It might be best to leave an additional comment here that this testing practice is generally not necessary (instead check the resource attribute directly) and only shown as an example, or choose an example like provider function testing which doesn't need a resource. 👍

helper/resource/testing.go
```diff
@@ -1612,6 +1921,88 @@ func testCheckResourceAttrPair(isFirst *terraform.InstanceState, nameFirst strin
...
+//		resource.Test(t, resource.TestCase{
+//			// Provider definition omitted.
+//			Steps: []resource.TestStep{
+//				{
+//					// Example resource containing a computed boolean attribute named "computed_attribute"
+//					Config: `resource "test_resource" "one" {}
+//
+//			        output test_resource_one_output {
+//			          value = test_resource.one
+//			        }
```
bflad, line 1993: Similar here regarding an additional code comment to discourage the practice, if its resource-based.

helper/resource/testing.go
```diff
@@ -1632,6 +2023,51 @@ func TestCheckOutput(name, value string) TestCheckFunc {
...
+//		resource.Test(t, resource.TestCase{
+//			// Provider definition omitted.
+//			Steps: []resource.TestStep{
+//				{
+//					// Example resource containing a computed string attribute named "computed_attribute"
+//					Config: `resource "test_resource" "one" {}
+//
+//							output test_resource_one_output {
+//								value = test_resource.one
+//							}`,
```
bflad, line 2058: Similar here regarding an additional code comment to discourage the practice, if its resource-based.

helper/resource/testing.go
```diff
@@ -1632,6 +2023,51 @@ func TestCheckOutput(name, value string) TestCheckFunc {
...
+//					ConfigStateChecks: []statecheck.StateCheck{
+//						statecheck.ExpectKnownOutputValueAtPath(
+//							"test_resource_one_output",
+//							tfjsonpath.New("computed_attribute"),
+//							StringContains("str")),
```
bflad, line 2063: Similar here regarding some form of built-in string regular expression check.

helper/resource/testing_sets.go
```diff
@@ -129,6 +171,57 @@ func TestCheckTypeSetElemNestedAttrs(name, attr string, values map[string]string
...
+//
+// Deprecated: State checks using [TestStep.Check] have been superseded by [TestStep.ConfigStateChecks].
+// Use the built-in [statecheck.ExpectKnownValue] state check, can be used in combination
+// with [knownvalue.ListExact], [knownvalue.ListPartial], [knownvalue.SetExact], or
+// [knownvalue.SetPartial], with nested [custom] [knownvalue.Check] instead.
+// TestMatchTypeSetElemNestedAttrs function will be removed in the next major version.
+//
+// The following is an example of using [statecheck.ExpectKnownValue] in combination
+// with [knownvalue.SetExact], with a nested [custom] [knownvalue.Check] to replicate
+// the behaviour of TestCheckTypeSetElemAttr.
```
bflad, line 183: Similar here regarding some form of built-in string regular expression check.

plancheck/expect_known_output_value_at_path_test.go
```diff
@@ -140,6 +140,100 @@ func TestExpectKnownOutputValueAtPath_CheckPlan_AttributeValueNull(t *testing.T)
...
+	t.Parallel()
+
+	r.Test(t, r.TestCase{
+		ProviderFactories: map[string]func() (*schema.Provider, error){
+			"test": func() (*schema.Provider, error) { //nolint:unparam // required signature
+				return testProvider(), nil
+			},
+		},
+		// Prior to Terraform v1.3.0 a planned output is marked as fully unknown
+		// if any attribute is unknown. The id attribute within the test provider
```
bflad, line 153: Since we know this information, it might be good to include it with the Go/website documentation surrounding output checks, if its not already. We'll want folks testing older versions of Terraform to know that outputs are fully unknown if any part of the value is unknown. 👍

website/docs/plugin/testing/acceptance-tests/known-value-checks/not-null.mdx
bflad, line 1: Certainly not saying its better, but another way to represent these checks that apply to any type would be bundling them into a single "any" or "generic" check page. NotNull could also live with Null since its just the opposite check. I guess we can wait to see how many pop up before trying to over-optimize the website documentation though. 😄

bflad (commented):
Hey @bendbennett 👋 I think this is getting pretty close. Please reach out with any questions/comments/concerns.

helper/resource/testing.go
```diff
@@ -1307,6 +1471,43 @@ func TestCheckResourceAttrWith(name, key string, checkValueFunc CheckResourceAtt
 // attributes using the special key syntax, checking a list, map, or set
 // attribute directly is not supported. Use TestCheckResourceAttr with
 // the special .# or .% key syntax for those situations instead.
+//
+// Deprecated: State checks using [TestStep.Check] have been superseded by [TestStep.ConfigStateChecks].
+// Use the built-in [statecheck.ExpectNoValueExists] instead.
+// TestCheckNoResourceAttr function will be removed in the next major version.
+//
+// The following is an example of using [statecheck.ExpectNoValueExists] to
+// replicate the behaviour of TestCheckNoResourceAttr.
```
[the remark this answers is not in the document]
bendbennett: I've removed `ExpectNoValueExists` and updated the docs to suggest using `ExpectKnownValue` with `knownvalue.Null` check type.

helper/resource/testing.go
```diff
@@ -1411,6 +1612,47 @@ func testCheckNoResourceAttr(is *terraform.InstanceState, name string, key strin
...
 // compilation.
+//
+// Deprecated: State checks using [TestStep.Check] have been superseded by [TestStep.ConfigStateChecks].
+// Use the built-in [statecheck.ExpectKnownValue] state check in combination
+// with a [custom] [knownvalue.Check] instead.
+// TestMatchResourceAttr function will be removed in the next major version.
+//
+// The following is an example of using [statecheck.ExpectKnownValue]
+// in combination with a [custom] [knownvalue.Check] to replicate the behaviour of
+// TestMatchResourceAttr.
```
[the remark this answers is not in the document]
bendbennett: Have added `knownvalue.StringRegularExpression`.

helper/resource/testing.go
```diff
@@ -1463,6 +1705,65 @@ func testMatchResourceAttr(is *terraform.InstanceState, name string, key string,
...
+//		testBool := Pointer(false)
+//
+//		resource.Test(t, resource.TestCase{
+//			// Provider definition omitted.
+//			Steps: []resource.TestStep{
+//				{
+//					// Example resource containing a computed boolean attribute named "computed_attribute"
+//					Config: `resource "test_resource" "one" {}`,
+//					ConfigStateChecks: []statecheck.StateCheck{
+//						AlterValue(testBool),
```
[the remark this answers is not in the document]
bendbennett: @bflad thank you for raising this.

Having dug into this a little further, I think it reveals that we need to introduce a new `CheckStateFunc` type and known value check types that accept a pointer, for the primitive forms of known value check types.

The following is included in this PR, but I wanted to lay it out here for clarity.

`CheckStateFunc` and the implementation of the `StateCheck` interface by the `CheckStateFunc` type look as follows:

```go
type CheckStateFunc func(context.Context, CheckStateRequest, *CheckStateResponse)

func (f CheckStateFunc) CheckState(ctx context.Context, req CheckStateRequest, resp *CheckStateResponse) {
	f(ctx, req, resp)
}
``` 

An example test, that uses an analogous flow to that which you described (thanks for the various links 👍), looks as follows:

```go
func TestExpectKnownValue_CheckState_BoolPointer(t *testing.T) {
	t.Parallel()

	var apiObj apiObject

	r.Test(t, r.TestCase{
		ProviderFactories: map[string]func() (*schema.Provider, error){
			"test": func() (*schema.Provider, error) { //nolint:unparam // required signature
				return testProvider(), nil
			},
		},
		Steps: []r.TestStep{
			{
				Config: `resource "test_resource" "one" {}`,
					makeAPICall("test_resource.one", &apiObj),
					statecheck.ExpectKnownValue(
						"test_resource.one",
						tfjsonpath.New("string_computed_attribute"),
						StringPtr(&apiObj.s),
					),
				},
			},
		},
	})
}

type apiObject struct {
	s string
}
```

The implementation of the `makeAPICall` function is as follows:

```go
func makeAPICall(name string, apiObj *apiObject) statecheck.CheckStateFunc {
	return func(ctx context.Context, req statecheck.CheckStateRequest, resp *statecheck.CheckStateResponse) {
		var resource *tfjson.StateResource

		if req.State == nil {
			resp.Error = fmt.Errorf("state is nil")

			return
		}

		if req.State.Values == nil {
			resp.Error = fmt.Errorf("state does not contain any state values")

			return
		}

		if req.State.Values.RootModule == nil {
			resp.Error = fmt.Errorf("state does not contain a root module")

			return
		}

		for _, res := range req.State.Values.RootModule.Resources {
			if name == res.Address {
				resource = res

				break
			}
		}

		if resource == nil {
			resp.Error = fmt.Errorf("%s - Resource not found in state", name)

			return
		}

		// Typically, an ID, or some other unique identifier would be extracted from the resource and 
		// supplied to an API client in order to make an API call to retrieve an API object.
		s, err := tfjsonpath.Traverse(resource.AttributeValues, tfjsonpath.New("string_computed_attribute"))

		if err != nil {
			resp.Error = err

			return
		}

		*apiObj = apiObject{s: s.(string)}
	}
}
```

The known value type for string pointer is as follows:

```go
var _ knownvalue.Check = stringPtr{}

type stringPtr struct {
	value *string
}

// CheckValue determines whether the passed value is of type string, and
// contains a matching string value.
func (v stringPtr) CheckValue(other any) error {
	otherVal, ok := other.(string)

	if !ok {
		return fmt.Errorf("expected string value for StringPtr check, got: %T", other)
	}

	if otherVal != *v.value {
		return fmt.Errorf("expected value %s for StringPtr check, got: %s", *v.value, otherVal)
	}

	return nil
}

// String returns the string representation of the string value.
func (v stringPtr) String() string {
	return *v.value
}

// StringPtr returns a Check for asserting equality between the
// supplied string and the value passed to the CheckValue method.
func StringPtr(value *string) stringPtr {
	return stringPtr{
		value: value,
	}
}
```

### Questions

If you're in agreement with the overall direction here, do you think we should:

- Add a convenience function which wraps the extraction of a resource attribute from state (e.g., id).
- Add known value check types for primitives that accept pointers.

Interested to hear your thoughts.

helper/resource/testing.go
bendbennett: `knownvalue.StringRegularExpression` has been added and the docs updated accordingly.

plancheck/expect_known_output_value_at_path_test.go
bendbennett: Have updated the Go docs and the website docs.

website/docs/plugin/testing/acceptance-tests/known-value-checks/not-null.mdx
bendbennett: Ok, will leave this as is for now. Happy to restructure going forwards if we feel like it makes the documentation clearer or we end up with a larger number of "generic" known value check types.

helper/resource/testing.go
austinvalle: > A note (that we don't necessarily need to solve in this PR, since it's already pretty large)

`knownvalue.StringRegularExpression` won't be a replacement for certain `TestMatchResourceAttr` checks that assumed the testing framework would treat a state value that is `json.Number` as a string. Like this example on the time provider (where hour, minute, etc. are all number values in state)

https://github.com/hashicorp/terraform-provider-time/blob/87a1ba82545baa44dc9c6770c4d5342aa617c7f2/internal/provider/resource_time_static_test.go#L26-L29

This seems more to the edge in-terms of use-case, but thought I would mention it as I just came across it.

bflad: My suggestion here is that it would be nice to support known value checks with the correct types and semantics rather than relying on certain implementation details that could change over time. We should encourage replacing workarounds like the one I had written there to ensure some level of full coverage. 😅 

More concretely, I would propose known value checks such as `Int64AtLeast(min int64)`, `Int64AtMost(max int64)`, `Int64Between(min, max int64)`, so this sort of checking would become `statecheck.ExpectKnownValue("time_static.test", tfjsonpath.New("day"), knownvalue.Int64Between(1, 31))`. 

There may also be a decent way to directly check via `time.Now()` as well by calling that in the test function and then passing its attribute values to checks, but I was worried about the slight time drift that will occur from when that function call occurs versus when the testing logic would actually plan/apply the resource.

austinvalle: I replaced the awkward `NumberRegularExpression` with a more reasonable `Int64Between` in that linked `time` provider PR: https://github.com/hashicorp/terraform-provider-time/commit/d73f9a9168e7dd1554fe1fbe2cb8f4f676405852

If we end up moving that into `terraform-plugin-testing` we can replace it later 👍🏻

helper/resource/testing.go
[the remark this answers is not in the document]
bendbennett: Following out-of-band discussion, it was decided that a built-in state check equivalent to `TestCheckResourceAttrPtr` would not be added in this PR. [Consider Multiple Value Checking Interface](https://github.com/hashicorp/terraform-plugin-testing/issues/295#top) outlines a potential alternative to `TestCheckResourceAttrPtr` which will be evaluated going forwards.

knownvalue/not_null.go
```diff
@@ -0,0 +1,32 @@
...
+	return nil
+}
+
+// String returns the string representation of notNull.
+func (v notNull) String() string {
+	return "not-null"
+}
+
+// NotNull returns a Check for asserting the value passed
+// to the CheckValue method is not nil.
```
austinvalle, line 29: Curious q: Does this require any sort of mentioning SDKv2 behavior of storing `""` vs `nil` and how `knownvalue.NotNull` won't detect that?

I don't fully understand the old SDKv2 behavior so this question may be off-base

knownvalue/string_regexp.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// String returns the string representation of the value.
+func (v stringRegularExpression) String() string {
+	return v.regex.String()
+}
+
+// StringRegularExpression returns a Check for asserting equality between the
+// supplied regular expression and a value passed to the CheckValue method.
+func StringRegularExpression(regex *regexp.Regexp) stringRegularExpression {
```
austinvalle, line 41: super nit 🦸🏻: Can we shorten this to `StringRegex`/`StringRegexp`?

The Go standard library already uses `regexp` and Terraform uses [`regex`](https://developer.hashicorp.com/terraform/language/functions/regex) so it seems reasonable to assume this wouldn't cause too much confusion.

website/docs/plugin/testing/acceptance-tests/known-value-checks/index.mdx
```diff
@@ -36,15 +36,17 @@ knownvalue.ListExact([]knownvalue.Check{
...
+| [Bool Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/bool)        | `schema.BoolAttribute`    | `schema.TypeBool`    |
+| [Float64 Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/float64)  | `schema.Float64Attribute` | `schema.TypeFloat`   |
+| [Int64 Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/int64)      | `schema.Int64Attribute`   | `schema.TypeInt`     |
+| [List Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/list)        | `schema.ListAttribute`    | `schema.TypeList`    |
+| [Map Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/map)          | `schema.MapAttribute`     | `schema.TypeMap`     |
+| [NotNull Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/not-null) | All                       | All                  |
+| [Null Known Value Checks](/terraform/plugin/testing/acceptance-tests/known-value-checks/null)        | All                       | All                  |
```
austinvalle, line 47: Similar question to above, are there caveats with using `Null` or `NotNull` with SDKv2 resource attributes?

website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
```diff
@@ -54,34 +64,142 @@ func TestExpectKnownOutputValue_CheckPlan_Bool(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		// Provider definition omitted.
+		Steps: []resource.TestStep{
+			{
+				// Example resource containing a computed boolean attribute named "computed_attribute"
+				Config: `resource "test_resource" "one" {}
+
+                // Generally, it is not necessary to use an output to test a resource attribute,
+                // the resource attribute should be tested directly instead. This is only shown as
+                // an example.
```
austinvalle, line 98: nit: Might be helpful to describe exactly what "directly" means in this context. (I assume this meant checking state with `statecheck.ExpectKnownValue`)

helper/resource/testing.go
```diff
@@ -1036,6 +1033,41 @@ func ComposeAggregateTestCheckFunc(fs ...TestCheckFunc) TestCheckFunc {
 // attributes using the special key syntax, checking a list, map, or set
 // attribute directly is not supported. Use TestCheckResourceAttr with
 // the special .# or .% key syntax for those situations instead.
+//
+// The following is an example of using [ExpectKnownValue] with [knownvalue.NotNull]
+// to replicate the behaviour of TestCheckResourceAttrSet.
```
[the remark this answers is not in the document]
bendbennett: I was thinking that the examples might serve for "organic" adoption of, or experimentation with, the new state checks rather than having an overt declaration that the `TestCheckFunc` implementations are legacy, or will be deprecated until such time as we're happy to move the new plan and state checks out of experimental status and have all of the alternative implementations for replicating built-in `TestCheckFunc`(s) in place. But I'm happy to add some additional verbiage if that's the consensus.

knownvalue/not_null.go
bendbennett: I'm a little reluctant to mention any SDK/v2-specific behaviour within the `knownvalue` package itself. Of course if the consensus is that we should make an explicit statement here regarding SDK/v2 behaviour then I'm happy to do that, but I'd be interested in hearing a little more detail around exactly what should appear in the Go docs under these circumstances. In terms of the behaviour of `knownvalue.NotNull` then it should be the same irrespective of whether SDK/v2 or the framework have been used to write the provider, as the check is just looking to see whether the value is not nil/null in the state or plan. 

The built-in state checks for `TestCheckResourceAttrSet`, and `TestCheckNoResourceAttr` are different in that null values are removed from the state that is retrieved and used for evaluating `TestCheckFunc`(s). For example, the [random provider](https://github.com/hashicorp/terraform-provider-random/blob/main/internal/provider/resource_string_test.go#L808), uses, [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckNoResourceAttr) to verify that an optional attribute (override_special) does not appear in state. This is a little peculiar, given that optional attributes that are not defined in configuration are stored in state as null. Investigation as to why this usage of [TestCheckNoResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckNoResourceAttr) executes successfully reveals that immediately prior to the TestStep.Check being run, the state is retrieved by calling [getState()](https://github.com/hashicorp/terraform-plugin-testing/blob/main/helper/resource/testing_new.go#L462). This calls shimStateFromJson(), and ultimately [shimmedFlatmap.AddEntry()](https://github.com/hashicorp/terraform-plugin-testing/blob/main/helper/resource/state_shim.go#L290) which omits adding an entry for the attribute in the returned terraform.State if the attribute is nil, which is the case if the attribute has been omitted from the configuration, or explicitly set to null in the configuration.

Hope I'm not off the mark with this. Let me know if we need to discuss this further.

knownvalue/string_regexp.go
bendbennett: Good point. Updated to use `StringRegexp`.

website/docs/plugin/testing/acceptance-tests/known-value-checks/index.mdx
bendbennett: As mentioned above, I'm a little reluctant to mention SDK/v2-specific issues in this context, but if there is consensus that this would add value then please feel free to suggest some verbiage to that effect.

website/docs/plugin/testing/acceptance-tests/plan-checks/output.mdx
bendbennett: I've updated the wording and add an example in each place where `// Generally, it is not necessary to use an output to test a resource....` appears.

knownvalue/not_null.go
austinvalle: Yeah that all makes sense to me. Probably best to avoid and if confusion arises, any documentation about `knownvalue` usage with SDKv2 would be best served in the SDKv2 section itself 😄

website/docs/plugin/testing/acceptance-tests/known-value-checks/index.mdx
austinvalle: Makes sense to me 👍🏻

helper/resource/testing.go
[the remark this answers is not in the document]
bflad: I lean a little towards Austin's clarity request here, it does feel like there needs to be something to describe why the example is there. Maybe something like "An experimental interface exists to to potentially replace this functionality in the future and feedback would be appreciated. This example performs the same check with that experimental interface:"

knownvalue/string_regexp.go
```diff
@@ -0,0 +1,45 @@
...
+// to StringRegexp.
+func (v stringRegexp) CheckValue(other any) error {
+	otherVal, ok := other.(string)
+
+	if !ok {
+		return fmt.Errorf("expected string value for StringRegexp check, got: %T", other)
+	}
+
+	if !v.regex.MatchString(otherVal) {
+		return fmt.Errorf("expected regex match %s for StringRegexp check, got: %s", v.regex.String(), otherVal)
```
bflad, line 28: Thank you for including the regex string in the error ❤️

website/docs/plugin/testing/acceptance-tests/known-value-checks/not-null.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+# NotNull Known Value Checks
+
+The known value checks that are available for values that are not null are:
+
+* [NotNull](/terraform/plugin/testing/acceptance-tests/known-value-checks/null#notnull-check)
+
+## `NotNull` Check
+
+The [NotNull](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/knownvalue#NotNull) check tests that a resource attribute, or output value is not null.
```
bflad, line 15: Nit: I wonder if its worth mentioning here or anywhere else that "not null" also means "any known value"?

bflad (approved):
Thanks for sticking through this -- looks good to me with some minor considerations on documentation 🚀

helper/resource/testing.go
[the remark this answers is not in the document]
bendbennett: Fair enough. Thanks for the suggested wording, have incorporated throughout.

website/docs/plugin/testing/acceptance-tests/known-value-checks/not-null.mdx
bendbennett: I've update the docs here. I didn't modify the Go docs for `knownvalue.NotNull`, as it seems explicit that the check is evaluating whether the passed value is nil or not.

PR #300, opened 2024-03-05
Bump Go version to 1.21

bendbennett:
Reference: https://github.com/hashicorp/terraform-providers-devex-internal/issues/175


bflad (approved):
Looks good to me 🚀

PR #304, opened 2024-03-08
tfversion: Add variable for version 1.9.0

bflad:
Terraform 1.8.0-beta1 has been released and a 1.9 release branch has already been cut with its own changelog entries so we know there will be 1.9 series. As with any of these `tfversion` version variable inclusions, it does not feel like it warrants a changelog entry for such as small change, but happy to create one.


bendbennett (approved):
LGTM 🚀

PR #305, opened 2024-03-11
Update `RandIpAddress()` so it can use any prefix as input

chrismarget:
The `RandIpAddress()` function has some confusing issues.

- The comment [`The prefix length must be less than 31`](https://github.com/hashicorp/terraform-plugin-testing/blob/147ecbbf9f4ceb71a922e91ccb16c42c96119b13/helper/acctest/random.go#L125) isn't correct. It will accept an IPv4 prefix with a 32-bit mask length, and IPv6 prefix bit masks must be at least 65 bits long. While I have doubts about the utility of "[give me a random value from within the range of *exactly one host*](https://www.youtube.com/watch?v=aUQkbXWwJhQ)", that functionality is validated by the tests.
- It won't, however, accept "0.0.0.0/0" (the shortest possible IPv4 prefix: "give me any address in the whole Internet"), but there doesn't seem to be any reason for this constraint.
- It also won't accept IPv6 prefixes shorter than /65 because of of how the random host portion is generated (only 63 bits available). Being unable to generate an IP within a /64 (standard subnet, let alone a reasonable sized block like the documentation prefix) is what led me to dig into this function.

This PR eliminates the confusing comment and lifts the unnecessary constraints for both protocols so that random addresses can be generated from IPv4 prefixes `0.0.0.0/0` through `255.255.255.255/32`, IPv6 prefixes `::/0` through `ff:ff:ff:ff:ff:ff:ff:ff/128`, and anywhere in between.


helper/acctest/random.go
```diff
@@ -134,47 +132,33 @@ func RandIpAddress(s string) (string, error) {
...
+		resultBytes[i] = (resultBytes[i] & inverseMaskBytes[i]) + prefixBytes[i]
 	}
 
-	// Calculate random address by taking prefix address and adding the random
-	// integer.
-	randAddrInt := new(big.Int).SetBytes(prefix.Addr().AsSlice())
-	randAddrInt.Add(randAddrInt, big.NewInt(randInt))
-
-	randAddr, ok := netip.AddrFromSlice(randAddrInt.Bytes())
+	result, _ := netip.AddrFromSlice(resultBytes)
```
austinvalle, line 159: nit: I know it's not possible at this point, given the logic above, but I feel like we should keep the original logic of checking the `ok` return from `netip.AddrFromSlice` and returning an error with the byte data:

```go
result, ok := netip.AddrFromSlice(resultBytes)
if !ok {
    return "", fmt.Errorf("unexpected error creating random address from bytes: %#v", resultBytes)
}
```

austinvalle (commented):
Overall I think these changes look good, I have one minor note and there are a couple small lint errors pointed out in the PR.

Also would you mind adding a [changelog](https://github.com/hashicorp/terraform-plugin-testing/blob/main/.github/CONTRIBUTING.md#new-pull-request) for this fix?

helper/acctest/random.go
```diff
@@ -134,47 +132,36 @@ func RandIpAddress(s string) (string, error) {
 		return prefix.Addr().String(), nil
 	}
 
-	prefixSizeExponent := uint(prefix.Addr().BitLen() - prefix.Bits())
-
-	if prefix.Addr().Is4() && prefixSizeExponent > 31 {
-		return "", fmt.Errorf("CIDR range is too large: %d", prefixSizeExponent)
+	// base address as byte slice
+	prefixBytes, err := prefix.Masked().Addr().MarshalBinary()
```
chrismarget, line 136: I've added `Masked()` here to ensure that a user asking for a random address within "192.168.1.77/24" isn't masking out random bits with that "77". They'll get any address within 192.168.1.0-255 this way.

helper/acctest/random.go
chrismarget: Hey @austinvalle,

I think I took care of your punch list.

Additionally, see the inline comment about the addition of `Masked()` (a change you didn't request, but seems like a better way to handle slightly-broken input).

Thanks!

helper/acctest/random.go
austinvalle: Ah! That makes sense. I can't really think of a scenario where you'd _want_ to lose that portion of the range, and honestly at that point, the cost of implementing your own random IP function is low 😄

austinvalle (approved):
Thanks for making those changes. Everything new looks good to me 🚀 

I have a slight addition to the changelog which I'll resolve and merge. Thanks for the contribution @chrismarget ! This will go out in the upcoming `v1.8.0` of the testing module 👍🏻

Issue #303, opened 2024-03-07
tfversion: Consider Ignoring Terraform Prerelease Version Information in Certain Checks


bflad:
### terraform-plugin-testing version

```
v1.7.0
```

### Use cases

When Terraform introduces new functionality, it will appear in prereleases of the minor version before general availability. This is true for provider-facing features, the Terraform built-in provider features, and purely configuration-based features.

The terraform-plugin-testing Go module supports automatic test skipping and test failures based on the Terraform version via the [`TestCase.TerraformVersionChecks` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCase.TerraformVersionChecks) and the various built-in check implementations in the [`tfversion` package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/tfversion).

In this example, tests will automatically be skipped if the Terraform version is below 1.8.0 (e.g. 1.7.x, 1.6.x, etc.):

```go
TerraformVersionChecks: []tfversion.TerraformVersionCheck{
	tfversion.SkipBelow(tfversion.Version1_8_0),
},
```

Especially when it comes to provider-facing features though, there is a time period during these Terraform prereleases where provider developers may want to implement the new features with acceptance testing, but also ensure that testing only runs on the upcoming version or later. For example, many of the HashiCorp-owned "utility" providers explicitly run their testing against every minor version of Terraform supported to detect potential regressions or version-specific quirks so these version checks working are important to not raise false positives or unexpectedly pass testing by skipping it.

Given the above check, running the testing while using Terraform 1.8.0-beta1 will skip the test:

```console
$ TF_ACC_TERRAFORM_PATH=/Users/bflad/bin/terraform1.8.0-beta1 TF_ACC=1 go test -count=1 -run='TestRFC3339Parse_UTC' -v ./internal/provider
=== RUN   TestRFC3339Parse_UTC
    function_rfc3339_parse_test.go:18: Terraform CLI version 1.8.0-beta1 is below minimum version 1.8.0: skipping test
--- SKIP: TestRFC3339Parse_UTC (1.18s)
```

For certain version checks, it may be desirable to "allow"/"ignore" prerelease versions given that the intention of these checks is to detect versions based on the core version information.

### Attempted solutions

As a workaround, it is possible to inject prerelease information into `tfversion` package checks:

```go
TerraformVersionChecks: []tfversion.TerraformVersionCheck{
[...]

PR #308, opened 2024-03-19
internal/testingiface: New testing.T equivalent interface and unit testing helpers

### LINKED ISSUE #254: Remove/Replace mitchellh/go-testing-interface Dependency

```
v1.6.0
```

### Use cases

https://github.com/mitchellh/go-testing-interface was archived earlier this year.

### Proposal

TBD

bflad:
Closes #254
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/303

_Note: This change is easiest to view with whitespace changes ignored._

As part of upcoming efforts to consider prerelease versioning more pragmatically in the `tfversion` package checks, the need to properly unit test `(*testing.T).Skip()` and friends behaviors is more apparent. Similar to issues with testing `(*testing.T).Fatal()` behaviors, the Go standard library in general is not the most friendly to capture test helper behaviors since the `testing.T` implementation is not intended to be extensible. This change was initially intended to extend our current usage of `github.com/mitchellh/go-testing-interface` as a stopgap, however while attempting to implement skip assertion testing, the `RuntimeT` implementation did not call `runtime.Goexit()` and would cause tests to continue to execute test logic after skipping even though in reality they would not. While the tests could be fixed to ensure they still passed due to that behavior change, it felt like a good time as any to replace the now-unsuitable and archived `github.com/mitchellh/go-testing-interface` with our own internal implementation.

The new `internal/testingiface.T` interface is a wholesale duplicate interface to the prior `github.com/mitchellh/go-testing-interface.T` interface. Swapping this interface in exported functions, such as `helper/resource.Test()` is only a compatibility issue for consumers if by chance they were implementing a wrapping function that also depended on `github.com/mitchellh/go-testing-interface.T` and passed that implementation detail directly to function calls of this Go module. This is a gray compatibility area as the intention of using `github.com/mitchellh/go-testing-interface.T` was for consumers to use real `*testing.T` implementations, which will work the same after this change, but it was not appropriately documented that extensions of exported functionality using `github.com/mitchellh/go-testing-interface.T` should not rely on that explicitly. Changing this detail is therefore a pragmatic compromise that it is more important to prioritize and document the intended usage while removing the archived dependency, which is almost exclusively how the ecosystem interacts with the functionality, over the potential of compatibility issues in a very small subset of the ecosystem.

Since `(*testing.T).Skip()` is considered a passing test, there are now assertions whether skipping was called or not, as expected, in the existing `tfversion` package tests. The new `internal/testingiface` package also includes similar assertion helpers for `(*testing.T).Fatal()` and `(*testing.T).Parallel()` for consistency and cleans up some existing logic.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/hashicorp/terraform-plugin-go v0.22.1
 	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.33.0
-	github.com/mitchellh/go-testing-interface v1.14.1
 	github.com/zclconf/go-cty v1.14.3
 	golang.org/x/crypto v0.21.0
 )
@@ -41,6 +40,7 @@ require (
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
+	github.com/mitchellh/go-testing-interface v1.14.1 // indirect
 	github.com/mitchellh/go-wordwrap v1.0.0 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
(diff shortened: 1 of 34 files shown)


bflad:
_Draft Note: This change warrants additional consideration and changelog entries, so leaving in draft for now._

PR #313, opened 2024-03-25
knownvalue: Add `TupleExact`, `TuplePartial` and `TupleSizeExact` checks for dynamic value testing

### LINKED ISSUE #312: Consider adding `Tuple*` known value checks

```
v1.7.0
```

### Proposal

Now that dynamic type support is in plugin framework, it's possible that provider developers will test dynamic data that could be a [tuple](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/tuple). For readability, we should introduce equivalent `TupleExact`, `TuplePartial`, and `TupleSize` known value checks.

### Workaround

Currently, you can test this with the `List*` equivalents, since no restriction is applied for `knownvalue` checks to be all the same type. Example with `ListExact`:

```go
func TestDynamicSchemaResource_DynamicAttribute(t *testing.T) {
    resource.UnitTest(t, resource.TestCase{
        ProtoV5ProviderFactories: map[string]func() (tfprotov5.ProviderServer, error){
            "framework": providerserver.NewProtocol5WithError(New()),
        },
        Steps: []resource.TestStep{
            {
                Config: `resource "framework_dynamic_schema" "test" {
                    dynamic_attribute = [true, 1234.5]
                }`,
                    statecheck.ExpectKnownValue("framework_dynamic_schema.test", tfjsonpath.New("dynamic_attribute"),
                        knownvalue.ListExact(
                            []knownvalue.Check{
                                knownvalue.Bool(true),
                                knownvalue.NumberExact(big.NewFloat(1234.5)),
                            },
                        ),
                    ),
                },
            },
        },
    })
}

```

austinvalle:
Closes #312

Exact copies of the `List*` equivalents.

### Notes

I explicitly didn't add documentation, since these are only really useful for dynamic value testing, but I can add them if we think it's necessary 👍🏻

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20240325-120539.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20240325-120539.yaml
@@ -1,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'knownvalue: Add `TupleExact`, `TuplePartial` and `TupleSizeExact` checks for
+  dynamic value testing.'
+time: 2024-03-25T12:05:39.777695-04:00
+custom:
+  Issue: "312"
--- a/knownvalue/list_partial_test.go
+++ b/knownvalue/list_partial_test.go
@@ -125,7 +125,7 @@ func TestListValuePartial_CheckValue(t *testing.T) {
 	}
 }
 
-func TestListValuePartialPartial_String(t *testing.T) {
+func TestListValuePartial_String(t *testing.T) {
 	t.Parallel()
 
 	got := knownvalue.ListPartial(map[int]knownvalue.Check{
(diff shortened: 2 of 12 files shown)


bflad (approved):
Looks good to me 🚀 

I think it might be worth the additional tuple documentation page, even if it is copy-paste-modify of the list page for now (maybe with a small callout that tuples are relatively uncommon except for testing dynamic types).

austinvalle:
> I think it might be worth the additional tuple documentation page, even if it is copy-paste-modify of the list page for now (maybe with a small callout that tuples are relatively uncommon except for testing dynamic types).

Added in 29d51b5 ⚡ - I adjusted the wording as well as added a little blurb mentioning they are only encountered when testing dynamic data 🚀

PR #323, opened 2024-04-16
workflows: Remove wildcard suffix from Terraform workflow call

austinvalle:
Can be merged after -> https://github.com/hashicorp/terraform-devex-repos/pull/69


SBGoods (approved):

austinvalle:
We had a test coverage file being generated with a filename from the terraform verison (which now has an `.*` in it 😆 )

Added a bash step to remove that wildcard before creating the file and we're all green now 👍🏻 🟢

Issue #295, opened 2024-02-25
Consider Multiple Value Checking Interface


bflad:
### terraform-plugin-testing version

```
1.6.0
```

### Use cases

There are certain patterns of provider acceptance testing which involving checking/comparing multiple values, typically computed resource attribute values from multiple `TestStep` states, including:

* Value Should Change: For example, ensuring that a computed attribute value is different after resource replacement, to assert that the resource replacement occurred as expected.
* Value Should Not Change: For example, ensuring that a computed attribute value is the same after resource update.
* Integer Value Should Increment: For example, ensuring that a computed attribute value, such as an API-handled resource-wide version number, increments after resource update/replacement. (Although now that plan checks exist for known values, this could be considered moot since checking that the resource generated the incremented known value during planning is just as valid/correct since Terraform will error if there is a mismatch after apply.)

When provider developers encounter this situation, there is not really a native/prescriptive way to handle the situation. It would be nice if at all possible to encourage developers towards more supported/prescriptive testing patterns. Having those patterns be baked into the provider acceptance testing API and extensible would better position us to document what developers should do or try in these situations, rather than potentially point at example code that may differ between providers and require extra explanation/re-familiarization each time.

### Attempted solutions

Overload the abstraction of the `StateCheckFunc`-based checking (`TestStep.Checks`) to extract API object types or resource attribute state values and potentially another function to implement value comparisons. Since there is no a canonical way to extract values out of API/state during each test step, developers are left to figure out this solution on their own. The additional "problem" with this setup is that it then is no longer a "check", even though much of the internal logic is the same as required for a state check to get to a resource attribute value.

From the hashicorp/aws provider, this example extracts API objects for comparison (which should instead be possible via state value comparison):

```go
// definitions

func testAccCheckBucketObjectExists(ctx context.Context, n string, v *s3.GetObjectOutput) resource.TestCheckFunc {
	return func(s *terraform.State) error {
		rs, ok := s.RootModule().Resources[n]
		if !ok {
			return fmt.Errorf("Not Found: %s", n)
		}

		conn := acctest.Provider.Meta().(*conns.AWSClient).S3Client(ctx)

		input := &s3.GetObjectInput{
			Bucket:  aws.String(rs.Primary.Attributes["bucket"]),
			Key:     aws.String(tfs3.SDKv1CompatibleCleanKey(rs.Primary.Attributes["key"])),
			IfMatch: aws.String(rs.Primary.Attributes["etag"]),
		}

[...]

Issue #282, opened 2024-01-25
Deprecation of Built-In TestCheckFunc Implementations for Checking Pairs of Values


bendbennett:
### Background

As part of the implementation of state checks which leverage [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State), built-in state checks, which are implementations of [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) (e.g., [TestCheckResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform/helper/resource#TestCheckResourceAttr)) are being deprecated in favour of state checks that are based on interrogating [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State). 

All of the implementations of [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) which handle checking individual values, will have an equivalent state check implemented with the new state checks that leverage [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State). However, the built-in implementations of [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckFunc) which deal with pairs of values will be deprecated without replacement.

| TestCheckFunc  | State Check |
| ------------- | ------------- |
| [TestCheckResourceAttrPair](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckResourceAttrPair)  | **Deprecated without replacement**  |
| [TestCheckTypeSetElemAttrPair](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/resource#TestCheckTypeSetElemAttrPair)  | **Deprecated without replacement** |

### Purpose of this Issue

The purpose of this issue is to provide a placeholder for signalling use-cases and interest in having built-in state checks which operate on [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State) that replicate the behaviour of `TestCheckResourceAttrPair` and/or `TestCheckTypeSetElemAttrPair`.

### References

- #266 
- #276

bflad:
[...]

Issue #187, opened 2023-09-12
New TestCheck helper `TestCheckResourceAttrPairWith`


ewbankkit:
### terraform-plugin-testing version

```
% go list -m github.com/hashicorp/terraform-plugin-testing/...
github.com/hashicorp/terraform-plugin-testing v1.5.1
```

### Use cases

It would be handy to have a new _TestCheck_ helper function that combines `TestCheckResourceAttrWith` and `TestCheckResourceAttrPair`, say `TestCheckResourceAttrPairWith`:

```go
type CheckResourceAttrPairWithFunc func(valueFirst, valueSecond string) error

func TestCheckResourceAttrPairWith(nameFirst, keyFirst, nameSecond, keySecond string, checkValuesFunc CheckResourceAttrPairWithFunc) TestCheckFunc {
...
}
```

The canonical use case for the Terraform AWS Provider is to test logical equivalence (rather than byte-for-byte equivalence) of a data source's IAM policy document and a resource's IAM policy document.

bendbennett:
The proposed `ValueComparer` interface in https://github.com/hashicorp/terraform-plugin-testing/issues/295 could be something that provides way to achieve the equivalent of `TestCheckResourceAttrPairWith` as a "custom" comparer could be supplied to a `CompareValuePairs` constructor. For example:

```go
func CompareValuePairs(resourceAddressOne string, attributePathOne tfjsonpath.Path, resourceAddressTwo string, attributePathTwo tfjsonpath.Path, comparer compare.ValueComparer) StateCheck {
	return &compareValuePairs{
		resourceAddressOne: resourceAddressOne,
		attributePathOne:   attributePathOne,
		resourceAddressTwo: resourceAddressTwo,
		attributePathTwo:   attributePathTwo,
		comparer:           comparer,
	}
}
```

PR #330, opened 2024-04-30
Implement comparison state checks to replace `TestCheckResourceAttrPtr`, `TestCheckResourceAttrPair` and `TestCheckTypeSetElemAttrPair`

bendbennett:
Closes: #295 
References: #282 
References: #187 

### Background

During the implementation of [Consider Adding State Check Equivalents to Plan Checks](https://github.com/hashicorp/terraform-plugin-testing/issues/266#top), it was intended that the pre-built `TestCheckFunc` implementations would be deprecated. However, these deprecations were not included at that time partly because there was no satisfactory re-implementation of `TestCheckResourceAttrPtr`, which used the new style state checks which leverage [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State), and neither `TestCheckResourceAttrPair`, nor `TestCheckTypeSetElemAttrPair` were re-implemented as it was unclear whether there was a need for such checks.

 ### Comparison State Checks and the `ValueComparer` interface

The concepts from [Consider Multiple Value Checking Interface](https://github.com/hashicorp/terraform-plugin-testing/issues/295#top) have been used to implement comparison state checks to replace the built-in `TestCheckFunc` checks for `TestCheckResourceAttrPtr`, `TestCheckResourceAttrPair`, and `TestCheckTypeSetElemAttrPair`. All three implementations make use of the `ValueComparer` interface.

#### Replacement for `TestCheckResourceAttrPtr` - `CompareValue`

```go
var _ StateCheck = &compareValue{}

type compareValue struct {
	resourceAddresses []string
	attributePaths    []tfjsonpath.Path
	stateValues       []any
	comparer          compare.ValueComparer
}

func (e *compareValue) AddStateValue(resourceAddress string, attributePath tfjsonpath.Path) StateCheck {
	e.resourceAddresses = append(e.resourceAddresses, resourceAddress)
	e.attributePaths = append(e.attributePaths, attributePath)

	return e
}

// CheckState implements the state check logic.
func (e *compareValue) CheckState(ctx context.Context, req CheckStateRequest, resp *CheckStateResponse) {
	/* ... */

	err = e.comparer.CompareValues(e.stateValues...)

	/* ... */
}

// CompareValue returns a state check that compares sequential values retrieved from state.
func CompareValue(comparer compare.ValueComparer) *compareValue {
	return &compareValue{
		comparer: comparer,
	}
}
```

#### Replacement for `TestCheckResourceAttrPair` - `CompareValuePairs`

```go
var _ StateCheck = &compareValuePairs{}

type compareValuePairs struct {
	resourceAddressOne string
	attributePathOne   tfjsonpath.Path
	resourceAddressTwo string
	attributePathTwo   tfjsonpath.Path
	comparer           compare.ValueComparer
}

// CheckState implements the state check logic.
func (e *compareValuePairs) CheckState(ctx context.Context, req CheckStateRequest, resp *CheckStateResponse) {
	/* ... */

	err = e.comparer.CompareValues(resultOne, resultTwo)

	/* ... */
}

func CompareValuePairs(resourceAddressOne string, attributePathOne tfjsonpath.Path, resourceAddressTwo string, attributePathTwo tfjsonpath.Path, comparer compare.ValueComparer) StateCheck {
	return &compareValuePairs{
		resourceAddressOne: resourceAddressOne,
		attributePathOne:   attributePathOne,
		resourceAddressTwo: resourceAddressTwo,
		attributePathTwo:   attributePathTwo,
		comparer:           comparer,
	}
}
```

#### Replacement for `TestCheckTypeSetElemAttrPair` - `CompareValueCollection`

```go
var _ StateCheck = &compareValueCollection{}

type compareValueCollection struct {
	resourceAddressOne string
	collectionPath     []tfjsonpath.Path
	resourceAddressTwo string
	attributePath      tfjsonpath.Path
	comparer           compare.ValueComparer
}

// CheckState implements the state check logic.
func (e *compareValueCollection) CheckState(ctx context.Context, req CheckStateRequest, resp *CheckStateResponse) {
	/* ... */

	for _, v := range results {
		switch resultTwo.(type) {
		case []any:
			errs = append(errs, e.comparer.CompareValues([]any{v}, resultTwo))
		default:
			errs = append(errs, e.comparer.CompareValues(v, resultTwo))
		}
	}

	/* ... */
}

// CompareValueCollection returns a state check that iterates over each element in a collection and compares the value of each element
// with the value of an attribute using the given value comparer.
func CompareValueCollection(resourceAddressOne string, collectionPath []tfjsonpath.Path, resourceAddressTwo string, attributePath tfjsonpath.Path, comparer compare.ValueComparer) StateCheck {
	return &compareValueCollection{
		resourceAddressOne: resourceAddressOne,
		collectionPath:     collectionPath,
		resourceAddressTwo: resourceAddressTwo,
		attributePath:      attributePath,
		comparer:           comparer,
	}
}
```

#### Implementations of the `ValueComparer` interface

```go
type ValueComparer interface {
	CompareValues(values ...any) error
}
```

```go
type ValuesSame struct{}

func (v ValuesSame) CompareValues(values ...any) error {
	for i := 1; i < len(values); i++ {
		if values[i-1] != values[i] {
			return fmt.Errorf("expected values to be the same, but they differ: %v != %v", values[i-1], values[i])
		}
	}

	return nil
}
```

There are implementations of the `ValueComparer` interface for:

- `ValuesSame`
- `ValuesDiffer`

These various implementations can be used in conjunction with `CompareValue`, `CompareValueCollection`, and `CompareValuePairs`.


compare/values_differ_any.go
```diff
@@ -0,0 +1,40 @@
...
+// CompareValues determines whether any value in the supplied values
+// is unique.
+func (v valuesDifferAny) CompareValues(values ...any) error {
+	if len(values) < 2 {
+		return nil
+	}
+
+	var val any
+
+	for i := 0; i < len(values); i++ {
```
austinvalle, line 24: nit: I might be oversimplifying this, but can we just take the first element and compare it to each of the remaining elements? Rather than the nested loop.

compare/values_differ_any.go
```diff
@@ -0,0 +1,40 @@
...
+		val = values[i]
+		for j := 1; j < len(values); j++ {
+			if !reflect.DeepEqual(values[i], values[j]) {
+				return nil
+			}
+		}
+	}
+
+	return fmt.Errorf("expected values to differ, but value is duplicated: %v", val)
```
austinvalle, line 33: nit: Should the error message indicate that it was expecting a single unique value? (i.e. all element values are %v)

compare/values_differ_any.go
bendbennett: Thinking about this a little further, I'm not sure there's a use case for `valuesDifferAny`. In fact, I'm not sure there's a use case for the built-in values comparers except for `valuesDiffer` and `valuesSame`. The 3 built-in compare value state checks are for comparing values retrieved from state from sequential `terraform apply` runs (i.e., `CompareValue` which replaces `TestCheckResourceAttrPtr`), comparing pairs of values retrieved from state (i.e., `CompareValuePairs` which replaces `TestCheckResourceAttrPair`), and determining if a collection contains a value (i.e., `CompareValueCollection` which replaces `TestCheckTypeSetElemAttrPair`). In each of these cases, the use cases for the value comparers is to determine whether the sequence of values passed in differ or are the same.

Consequently, I'm going to remove `valuesDifferAll`, `valuesDifferAny`, and `valuesSameAny`.

compare/values_differ_any.go
bendbennett: As above, I'm going to remove `valuesDifferAll`, `valuesDifferAny`, and `valuesSameAny`.

compare/values_differ_any.go
austinvalle: Ah okay, that makes sense. We could always come back and add more comparer implementations if necessary 👍🏻

SBGoods (approved):
LGTM!

PR #331, opened 2024-05-03
all: Add deferred action testing support (plan checks, version check, and CLI options)

austinvalle:
This PR adds a couple different things all towards supporting testing deferred actions:
- Two plan checks that enable inspecting the new deferred changes field in the plan JSON, ref: https://github.com/hashicorp/terraform-json/pull/123
- A new version check that ensures a test is only run on pre-release builds of Terraform
- A new `TestCase` field for enabling a limited set of flags, in this scenario, the `-allow-deferral` flag, ref: https://github.com/hashicorp/terraform-exec/pull/447
- I also updated the `terraform-plugin-go` version to `v0.23.0` to help with testing the new plan checks, ref: https://github.com/hashicorp/terraform-plugin-go/pull/403
   - This required updating some of the `testprovider` to implement the now required methods from `terraform-plugin-go`


plancheck/expect_deferred_change_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func Test_ExpectDeferredChange_Reason_Match(t *testing.T) {
+	t.Parallel()
+
+	r.Test(t, r.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.SkipBelow(tfversion.Version1_9_0),
+			tfversion.SkipIfNotPrerelease(),
```
austinvalle, line 27: I'm going to introduce a change to our terraform repo to ensure this CI actually has an alpha version to run these 😆

austinvalle: https://github.com/hashicorp/terraform-devex-repos/pull/71

austinvalle: Added, `terraform-plugin-testing` tests are running on `1.9.0-alpha20240501` - 
https://github.com/hashicorp/terraform-plugin-testing/actions/runs/8944501602/job/24635889715

go.mod
```diff
@@ -13,9 +13,9 @@ require (
 	github.com/hashicorp/hc-install v0.6.4
 	github.com/hashicorp/hcl/v2 v2.20.1
 	github.com/hashicorp/logutils v1.0.0
-	github.com/hashicorp/terraform-exec v0.20.0
-	github.com/hashicorp/terraform-json v0.21.0
-	github.com/hashicorp/terraform-plugin-go v0.22.2
+	github.com/hashicorp/terraform-exec v0.20.1-0.20240501214311-a1abb29f777f
+	github.com/hashicorp/terraform-json v0.21.1-0.20240502194054-3b8a921d9133
+	github.com/hashicorp/terraform-plugin-go v0.22.3-0.20240503133000-f7070c19a38c
```
austinvalle, line 18: TODO: update when `terraform-plugin-go`, `terraform-json` and `terraform-exec` are released

plancheck/expect_no_deferred_changes.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if !req.Plan.Complete {
+		resp.Error = errors.New("expected plan to be marked as complete, but complete was \"false\", indicating that at least one more plan/apply round is needed to converge.")
+		return
+	}
+}
+
+// ExpectNoDeferredChanges returns a plan check that asserts that there are no deffered changes
+// for any resources in the plan.
+func ExpectNoDeferredChanges() PlanCheck {
```
SBGoods, line 45: Would it be useful to have a `ExpectNoDeferredChange` plan check assertion for a single resource?

SBGoods (approved):
LGTM! Only one non-blocking question about a potential new check.

plancheck/expect_no_deferred_changes.go
austinvalle: Good question! Off the top of my head I can't think of a specific testing flow that would need a negative check like that.

My initial thought is that most tests would follow the general flow of testing a single resource:
- Test step 1: Check a resource is causing a plan to be deferred
- Test step 2: Check there are no more deferred changes (i.e. no more rounds needed)

I'm not sure if there is a workflow where a test may want to end before all deferred changes are satisfied, but maybe more provider developer usage will let us know if that'd be beneficial.

bendbennett (approved):
LGTM 🚀

go.mod
austinvalle: This has been updated with the stable releases 🎉

PR #388, opened 2024-11-08
tfversion: Add `SkipIfNotAlpha` version check

austinvalle:
Since we've recently added beta prereleases to our acceptance testing, we started getting test failures for experimental features like deferred actions that use the `SkipIfNotPrerelease` version check. That's because the only prerelease builds that get experimental features (like deferred actions) are `alpha` and source builds.

None of our CI uses source builds, so I just created a new `SkipIfNotAlpha` version check we can use for experimental features, while allowing `beta` builds to run in our CI successfully 👍🏻

### DIFF
--- a/.changes/unreleased/FEATURES-20241108-160634.yaml
+++ b/.changes/unreleased/FEATURES-20241108-160634.yaml
@@ -1,0 +1,6 @@
+kind: FEATURES
+body: 'tfversion: Added `SkipIfNotAlpha` version check for testing experimental features
+  of alpha Terraform builds.'
+time: 2024-11-08T16:06:34.662822-05:00
+custom:
+  Issue: "388"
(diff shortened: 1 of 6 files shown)


austinvalle:
You can see some example test failures in our recent dependabot PRs when it fails to skip properly for `1.10.0-beta1`: https://github.com/hashicorp/terraform-plugin-testing/actions/runs/11742344241/job/32712881987?pr=387

```bash
--- FAIL: Test_ExpectDeferredChange_Reason_Match (0.51s)
    expect_deferred_change_test.go:24: Step 1/1 error: Error running pre-apply plan: -allow-deferral is only
available in experimental Terraform builds: experiments are not enabled in version 1.10.0-beta1, as it's not
an alpha or dev build
```

PR #389, opened 2024-11-08
echoprovider: Add a new test-only v6 provider that echoes ephemeral provider data

austinvalle:
> The changes in this PR can be seen in use over at: https://github.com/hashicorp/terraform-provider-corner/pull/283

This PR contains an implementation of a v6 Terraform provider that can be used to test [ephemeral resources](https://developer.hashicorp.com/terraform/plugin/framework/ephemeral-resources). As ephemeral data is not stored in Terraform artifacts (plan or state), that means that the `terraform-plugin-testing` Go module cannot help provider developers test that data directly.

The `echo` provider utilizes the ephemeral nature of provider configurations as a "passthrough" for data to be echoed later by the single `echo` managed resource type:

```terraform
ephemeral "examplecloud_thing" "one" {
  name = "John Doe"
}

provider "echo" {
  # Provide the ephemeral data we want to run test assertions against
  data = ephemeral.examplecloud_thing.one
}

# The ephemeral data will be echoed into state
resource "echo" "test_one" {}

output "name_out" {
  # Will output "John Doe"!
  value = echo.test_one.data.name
}	
```

A more detailed example, along with accompanying state checks, can be found in the PRs website documentation: https://github.com/hashicorp/terraform-plugin-testing/blob/av/ephemeral-values/website/docs/plugin/testing/acceptance-tests/ephemeral-resources.mdx#testing-ephemeral-data-with-echo-provider

-----------

Eventually, other methods of inspecting/consuming ephemeral data may be introduced that might warrant revisiting this provider, such as:
- Terraform core implementing their own "debug" utility for practitioners inspecting ephemeral data
- Write-only attributes in managed resources

But for now, `ephemeral data => provider config => managed resource state` is essentially the only reasonable option for CI testing. (outside of using the remote system, which is encouraged in the docs) 

## Notes
- This PR also contains a new version constraint for the upcoming Terraform `v1.10.0` release and an implementation of `EphemeralResourceServer` RPCs as mentioned in https://github.com/hashicorp/terraform-plugin-testing/pull/382#issuecomment-2449846184
- `echoprovider` is implemented as a v6 provider because it's intended use is with ephemeral data, which exists only in Terraform v1.10.0+


echoprovider/server.go
```diff
@@ -0,0 +1,405 @@
...
+			PlannedState: req.ProposedNewState,
+		}, nil
+	}
+
+	// If we're updating and the provider config data hasn't changed, return prior state to indicate no diff
+	if priorState.Equal(tftypes.NewValue(echoTestSchema.ValueType(), map[string]tftypes.Value{"data": e.providerConfigData})) {
+		return &tfprotov6.PlanResourceChangeResponse{
+			PlannedState: req.PriorState,
+		}, nil
+	}
```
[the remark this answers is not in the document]
austinvalle: This would essentially make the resource immutable, so the only thing you could do is create/destroy it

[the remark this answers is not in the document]
austinvalle: I've gone ahead and made those changes here, along with adding a section of caveats: e5b632d36863fa4fe930bce9c937906e2f2d6757

website/docs/plugin/testing/acceptance-tests/ephemeral-resources.mdx
```diff
@@ -0,0 +1,274 @@
...
+
+</Highlight>
+
+# Ephemeral Resources
+
+[Ephemeral Resources](/terraform/language/v1.10.x/resources/ephemeral) are an abstraction that allows Terraform to reference external data, similar to [data sources](/terraform/language/data-sources), without persisting that data to plan or state artifacts. The `terraform-plugin-testing` module exclusively uses Terraform plan and state artifacts for it's assertion-based test checks, like [plan checks](/terraform/plugin/testing/acceptance-tests/plan-checks) or [state checks](/terraform/plugin/testing/acceptance-tests/state-checks), which means that ephemeral resource data cannot be asserted using these methods alone.
+
+In addition to this, ephemeral resources have a different [lifecycle](/terraform/language/v1.10.x/resources/ephemeral#lifecycle) then resources or data sources and must be [referenced](/terraform/language/v1.10.x/resources/ephemeral#referencing-ephemeral-resources) as a dependency to a managed resource or data source to be invoked. This means for acceptance testing ephemeral resource logic, provider developers should use a real world example where the ephemeral resource would be referenced.
```
austinvalle, line 17: TODO: This wording needs to be adjusted. I can't remember if this used to be the case, or if my brain fabricated it, but you definitely can just put an ephemeral resource in a config and run it, and it will execute `Open`

austinvalle: Done ⚡ 4644758b79e5e1f91bd5ff65061f43e0076fce2e

I absolutely fabricated this 😆

echoprovider/server_test.go
SBGoods, line 1: Do we want to add a test case showing the immutability of a single echo resource (ie showing that prior state is preserved between separate plan and apply cycles).

austinvalle: Good idea! Here is one -> 10123e58398933cf34f7e5201d7e418d7eca3bfe ⚡

SBGoods (approved):

PR #431, opened 2025-02-11
What if ... ImportBlockWithResourceIdentity

bbasata:
Cross-ref https://github.com/hashicorp/terraform-plugin-testing/pull/141

cc: @austinvalle

PR #441, opened 2025-03-05
Add additional import test modes 

stephybun:
Based off of https://github.com/hashicorp/terraform-plugin-testing/pull/431

TODOs:

- [ ] Add additional test cases e.g. for `ImportStateVerifyIgnore`, `SkipDataSource`, `ExpectError`?

### DIFF
--- a/helper/resource/testing_new.go
+++ b/helper/resource/testing_new.go
@@ -133,7 +133,7 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
 
 	// use this to track last step successfully applied
 	// acts as default for import tests
-	var appliedCfg teststep.Config
+	var appliedCfg string
 	var stepNumber int
 
 	for stepIndex, step := range c.Steps {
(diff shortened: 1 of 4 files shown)


helper/resource/testing_new.go
```diff
@@ -436,18 +436,6 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
...
-				File:      step.ConfigFile,
-				Raw:       mergedConfig,
-				TestStepConfigRequest: config.TestStepConfigRequest{
-					StepNumber: stepIndex + 1,
-					TestName:   t.Name(),
-				},
-			}.Exec()
-
-			appliedCfg = teststep.Configuration(confRequest)
-
```
stephybun, line 450: It looks like this was only ever needed and done for the import tests, so I've opted to move this into `testing_new_import_state.go` and pass the `appliedConfig` through as a string so it can be easily appended/modified based on the import mode/kind we're using in the test.

stephybun:
Replaced by #442

PR #442, opened 2025-03-05
Add additional import test modes

stephybun:
Based off of https://github.com/hashicorp/terraform-plugin-testing/pull/431
Supersedes #441

TODOs:

- [x] Add additional test cases e.g. for `ImportStateVerifyIgnore`, `SkipDataSource`?
- [ ] Add logic to compare the import plan with the state from the previous test step https://github.com/hashicorp/terraform-plugin-testing/blob/805da5ac7f2ec4553a24e2a8b7b8f6adff98faec/helper/resource/testing_new_import_state.go#L216
- [ ] Consider splitting this new import testing mode out into it's own file e.g. `testing_new_plannable_import.go`
- [ ] Enable these new import modes for configs supplied via file or directory
- [ ] Automatically add the terraform block to in-lined configs in the import step https://github.com/hashicorp/terraform-plugin-testing/blob/1b4767178c4323eeadfaa4a7785f9fc8343dddd7/helper/resource/testing_new_import_block_id_test.go#L445-L451


helper/resource/testing_new.go
```diff
@@ -436,18 +436,6 @@ func runNewTest(ctx context.Context, t testing.T, c TestCase, helper *plugintest
...
-				File:      step.ConfigFile,
-				Raw:       mergedConfig,
-				TestStepConfigRequest: config.TestStepConfigRequest{
-					StepNumber: stepIndex + 1,
-					TestName:   t.Name(),
-				},
-			}.Exec()
-
-			appliedCfg = teststep.Configuration(confRequest)
-
```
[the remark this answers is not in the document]
bbasata: 💯

helper/resource/testing_new_import_state.go
```diff
@@ -93,11 +93,35 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+			import {
+				to = %s
+				id = %q
+			}
+		`, step.ResourceName, importId)
+		default:
+			// Not an import block test so nothing to do here
+		}
+
+		confRequest := teststep.PrepareConfigurationRequest{
```
bbasata, line 114: I wonder how this interacts with the earlier `teststep.PrepareConfigurationRequest` from `step.Config` on line 26. Would this change the behavior of any existing `ImportCommandWithId` tests that define `step.Config`? I actually don't know 😃

helper/resource/testing_new_import_state.go
```diff
@@ -93,11 +93,35 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+			Directory: step.ConfigDirectory,
+			File:      step.ConfigFile,
+			Raw:       cfgRaw,
+			TestStepConfigRequest: config.TestStepConfigRequest{
+				StepNumber: stepIndex + 1,
```
bbasata, line 119: Something I had overlooked in the proof-of-concept pull request: what is the significance of `stepIndex + 1` here (and at line 26)?

stephybun: I believe it's to maintain the correct step number which was modified to index starting from 1 for a better UX

https://github.com/hashicorp/terraform-plugin-testing/blob/7e59fd9813b7404cc5a55ad58a757f3b3bb030e5/helper/resource/testing_new.go#L144

This instance (as well as another in `testing_new.go`) can probably be updated to use `stepNumber`

helper/resource/testing_new_import_state.go
stephybun: I can confirm that this doesn't change the behaviour of existing `ImportCommandWithId` tests that define a `Config` in-line.

helper/resource/testing_new_import_block_id_test.go
```diff
@@ -0,0 +1,459 @@
...
+					}
+
+					return nil
+				},
+			},
+		},
+	})
+}
+
+func TestTest_TestStep_ImportBlockId_ImportStateVerifyIgnore_Real_Example(t *testing.T) {
```
stephybun, line 214: On the fence about how useful this actually is for testing the behaviour of `ImportStateVerifyIgnore`. Considering removing it before this goes in.

helper/resource/testing_new_import_state.go
```diff
@@ -93,11 +93,35 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+		switch step.ImportStateKind {
+		case ImportBlockWithResourceIdentity:
+			t.Fatalf("TODO implement me")
+		case ImportBlockWithId:
+			cfgRaw += fmt.Sprintf(`
+			import {
+				to = %s
+				id = %q
+			}
+		`, step.ResourceName, importId)
```
SBGoods, line 109: We are only adding the import block to the config automatically if the `ImportBlockWithId` mode is set in a step without a config, is this expected? I wonder if we should always add the import block automatically to the config for `ImportBlockWithId` kinds. 

Currently, you can run a test with `ImportBlockWithId` set with a config that does not have an import block:
```
{
  Config: `
  terraform {
  	  required_providers {
    	  examplecloud = {
      		  source = "registry.terraform.io/hashicorp/examplecloud"
    	  }
	  }
  }
  
  resource "examplecloud_container" "test" {
	  name = "somename"
	  password = "somevalue"
  }`,
ResourceName:            "examplecloud_container.test",
ImportState:             true,
ImportStateKind:         ImportBlockWithId,
ImportStateVerify:       true,
ImportStateVerifyIgnore: []string{"password"},
},
```
With the current implementation, this would be running a regular import test (not a block id test) but there is no indication to provider developers that that is the case.

stephybun: That is a good point, my thinking was purely based on how import tests have been used in the azurerm provider which is that we _always_ re-use the config in the initial create test step so there is never need to specify `Config` for an import test. That assumption has been blown completely out of the water after looking into how tests using an import block function. Either way you're right that an import block should always be appended to the config regardless of where the it's coming from if `ImportBlockWithId` has been specified.

bbasata (approved):
This is very ready to merge 😃 

* Feature toggle: the PR adds a work-in-progress, optional pre-release feature; this functionality is only activated when the test author sets a non-zero (i.e. non-default) value on `TestStep.ImportStateKind`.

* Regression-safety: the refactoring to support the new feature looks safe, and this module's tests are passing.

We'll iterate on this feature in follow-up PRs.

This was a joy, @stephybun ✨

PR #461, opened 2025-03-21
importstate: a working example of ImportState as the first TestStep

bbasata:
This PR adds a test to resolve the question: what happens if the first `TestStep` uses import mode?

This test passes 🍏 

Notably: `ImportStateVerify` does not work, because there is no previous step state to verify against.


helper/resource/importstate/import_as_first_step_test.go
```diff
@@ -0,0 +1,65 @@
...
+			"examplecloud": providerserver.NewProviderServer(testprovider.Provider{
+				Resources: map[string]testprovider.Resource{
+					"examplecloud_container": examplecloudResource(),
+				},
+			}),
+		},
+		Steps: []r.TestStep{
+			{
+				ResourceName:    "examplecloud_container.test",
+				ImportStateId:   "examplecloud_container.test",
```
bbasata, line 34: Without this, `ImportState` will error when it attempts to extract an ID from previous test step state.

helper/resource/testing_new_import_state.go
```diff
@@ -238,8 +241,18 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
 		// TODO compare plan to state from previous step
+		err = runProviderCommand(ctx, t, func() error {
+			var err error
+			logging.HelperResourceDebug(ctx, "Run terraform apply")
+			err = importWd.Apply(ctx)
+			return err
+		}, importWd, providers)
+		if err != nil {
+			return err
+		}
```
austinvalle, line 252: 🤔 I'm still not sure we should actually apply anything during `ImportState` tests. It feels like an easy way to mess up remote infrastructure during a test without the expectations that a test assertion will point out what went wrong.

The logic above checks the first import plan to be a no-op, but it's still possible for a refresh to change what this apply does with no real feedback to the user.

------

The consequence of not running apply is that we can't use other test steps after an `ImportState`, but maybe that's okay? If the provider developer wants to run an actual apply, they could always use `Config` mode, so maybe that's where we draw the line?

helper/resource/importstate/import_block_as_first_step_test.go
```diff
@@ -0,0 +1,68 @@
...
+					}
+					if states[0].Attributes["location"] != "westeurope" {
+						return fmt.Errorf("unexpected location: %s", states[0].Attributes["location"])
+					}
+					return nil
+				},
+			},
+			{
+				RefreshState: true,
+				Check:        r.TestCheckResourceAttr("examplecloud_container.test", "name", "somevalue"),
```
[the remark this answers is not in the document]
bbasata: Got it.

`ConfigStateChecks` seems not to work with `RefreshState`. tells me that it requires `Config`, `ConfigFile`, or `ConfigDirectory`.

I'm inclined to remove the `RefreshState` step entirely - the `ImportStateCheck` in Step 1 tells us what we want to know.

helper/resource/testing_new_import_state.go
bbasata: I've made these changes based on this feedback:

* Removed the apply logic
* Added support for exactly one flavor of `ImportPlanChecks`: `PreApply`

bbasata:
> I'm still not sure we should actually apply anything during `ImportState` tests.

One detail that I remembered: the only reason `RefreshState` worked here is `ImportStatePersist`. Otherwise, `ImportState` operates in an isolated working directory.

As mentioned above, I've removed `RefreshState` from the tests, so I can also remove `ImportStatePersist` from the tests for now.

helper/resource/testing.go
```diff
@@ -810,6 +817,13 @@ type ConfigPlanChecks struct {
 	PostApplyPostRefresh []plancheck.PlanCheck
 }
 
+// ImportPlanChecks defines the different points in an Import TestStep when plan checks can be run.
+type ImportPlanChecks struct {
+	// PreApply runs all plan checks in the slice. This occurs after the plan of an Import test is computed. This slice cannot be populated
+	// with TestStep.PlanOnly, as there is no PreApply plan run with that flag set. All errors by plan checks in this slice are aggregated, reported, and will result in a test failure.
+	PreApply []plancheck.PlanCheck
```
austinvalle, line 824: 🤔 - Wondering if we can find a better name for this field. The naming suggests that we will apply the plan after running the checks.

One of the reason that `ConfigPlanChecks` naming is so awkward is because there were so many places to run plan checks that we basically needed to describe the location of where the plan checks executed (pre-apply, post-apply, etc.)

Here, if we're only going to run the plan checks once, maybe we don't need a separate struct at all? Maybe we could following `ConfigStateChecks` which is just a single slice of state checks that runs once.

helper/resource/testing_new_import_state.go
```diff
@@ -237,8 +240,14 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 		}
 
 		// TODO compare plan to state from previous step
+
+		if err := runPlanChecks(ctx, t, plan, step.ImportPlanChecks.PreApply); err != nil {
+			return err
+		}
+
```
austinvalle, line 247: Not directly related to this PR, but I noticed that we don't immediately return after running this import/plan and we probably should? Or wrap the remaining of the logic in this function in the following else block.

helper/resource/testing_new_import_state.go
```diff
@@ -237,8 +240,14 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 		}
 
 		// TODO compare plan to state from previous step
+
+		if err := runPlanChecks(ctx, t, plan, step.ImportPlanChecks.PreApply); err != nil {
+			return err
```
austinvalle, line 245: nit: Should we wrap the error messages produced by the plan checks for readability?

```go
    return fmt.Errorf("Import plan check(s) failed:\n%w", err)
```

helper/resource/importstate/import_block_as_first_step_test.go
```diff
@@ -0,0 +1,58 @@
...
+				},
+			}),
+		},
+		Steps: []r.TestStep{
+			{
+				ResourceName:    "examplecloud_container.test",
+				ImportStateId:   "examplecloud_container.test",
+				ImportState:     true,
+				ImportStateKind: r.ImportBlockWithId,
+				// ImportStateVerify: true,
```
austinvalle, line 41: I think this is already on your radar, but we should consider adding some explicit validation to TestStep to make it clear which fields you're allowed to use / will work together.

is this commented out to show that it's not possible to use here?

helper/resource/testing_new_import_state.go
bbasata: 💯

helper/resource/importstate/import_block_as_first_step_test.go
bbasata: > we should consider adding some explicit validation to TestStep to make it clear which fields you're allowed to use / will work together

💯. An upcoming pull request.

> is this commented out to show that it's not possible to use here?

This comment disappears in the PRs stacked atop this one, where we introduce `ImportPlanVerify`.

helper/resource/testing_new_import_state.go
bbasata: ✅ Noted for a follow-up PR -- it will be easier to see the current picture after we merge this & the two PRs "stacked" atop it.

helper/resource/testing_new_import_state.go
bbasata: ✅ Noted for a follow-up PR -- it will be easier to see the current picture after we merge this & the two PRs "stacked" atop it.

helper/resource/testing.go
bbasata: > I like the last two the most, `PreImport` and `PreApply`, and now I'm leaning back towards `PreApply`

Let's :shipit: with `PreApply` and use the upcoming prerelease as a vehicle for feedback.

austinvalle: Sounds good to me 👍🏻

PR #463, opened 2025-03-22
Nice debug logs

bbasata:
This PR aims for visually-scannable debug output from tests.

It changes the detailed logging of the provider lifecycle in `resource.runProviderCommand` from DEBUG to TRACE level. I have used this patch locally for nearly a week. In one example, debug output improves [from 166 lines to 30 lines](https://gist.github.com/bbasata/72a70ed83ade2bb36d4db107a604e888).

The DEBUG-level logging originated 3 years ago in https://github.com/hashicorp/terraform-plugin-testing/commit/b5c80aa8ad22a69562d30e2e7d1ed56831b5aebf. I feel the provider lifecycle is stable enough that we can change it to TRACE logging. `resource.runProviderCommand` has logic to handle various provider errors by returning an `error`.

### DIFF
--- a/.changes/unreleased/NOTES-20250325-084449.yaml
+++ b/.changes/unreleased/NOTES-20250325-084449.yaml
@@ -1,0 +1,5 @@
+kind: NOTES
+body: reduced the volume of DEBUG-level logging to make it easier to visually scan debug output
+time: 2025-03-25T08:44:49.949718-04:00
+custom:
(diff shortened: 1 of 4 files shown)


rainkwan:
Seems like a good idea so long as there wasn't anyone relying on the old logging output to debug in some way. I assume not though, should still provide the same information I would think.

PR #467, opened 2025-03-25
importstate: compare plan to previous state

bbasata:


helper/resource/testing.go
```diff
@@ -700,6 +700,23 @@ type TestStep struct {
...
+	// (ImportBlockWithID or ImportBlockWithResourceIdentity) agrees with
+	// the known state of a previous test stepstate values
+	// that are finally put into the state after import match for all the
+	// IDs returned by the Import.  Note that this checks for strict equality
+	// and does not respect DiffSuppressFunc or CustomizeDiff.
+	//
+	// By default, the prior resource state and import resource state are
+	// matched by the "id" attribute. If the "id" attribute is not implemented
+	// or another attribute more uniquely identifies the resource, set the
+	// ImportStateVerifyIdentifierAttribute field to adjust the attribute for
```
ansgarm, line 713: The PR description states that `ImportStateVerifyIdentifierAttribute` is not mirrored – am I missing something or would this comment need an update?

bbasata: The comment needs an update. Nice catch 😃 I've pushed a commit to change this.

internal/teststep/verify_import_plan_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"testing"
+
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-testing/internal/teststep"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+func Test_VerifyImportPlan(t *testing.T) {
```
austinvalle, line 14: One nice thing for this test would be to expand it for the other types of data outside of primitives (lists, sets, maps, single nested object/block, list nested objects/blocks, set nested objects/blocks, map nested attribute)

austinvalle (commented):
Looking good so far! I left some comments to consider 👀

helper/resource/testing_new_import_state.go
```diff
@@ -72,7 +72,7 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 	}
 
 	// Determine the ID to import
-	var importId string
+	var importId string //nolint:revive
```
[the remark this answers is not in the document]
bbasata: To make a long story short: I was seeing constant reminders that this is `Id` and not `ID`, and I've tracked it down to an editor plugin that was adding `--enable revive --exclude-use-default=false` to `golangci-lint`.

I've finally found the right combination of options & patches to turn these off. So I will remove this code comment.

The rule in effect here was the `var-naming` rule in `revive`: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming. 

It reports ~14 lines in this repo with either `Id` (it prefers `ID`) or `Json` (`JSON`). Nearly half of them are exported identifiers.

I'm +1 on linter improvements ... some day! 😃

bbasata:
Closing in favor of #476.

PR #476, opened 2025-04-04
helper/resource: plannable import re-work to support resources with dependencies

bbasata:
This change re-works plannable import testing to support a scenario uncovered in alpha releases: the resource-to-be-imported requires replacement ... because it depends on a resource that does not exist in state before the plan is generated.

The new approach: a plannable import `TestStep` constructs this state: `[copy of state from previous config mode TestStep] minus [the "resource under test," indicated by TestStep.ResourceName]`.

https://github.com/hashicorp/terraform-plugin-testing/blob/88a9bc9b2c440b6e0f632721e0ca69f583a13c28/helper/resource/testing_new_import_state.go#L155-L169

The plan is generated against this prior state; therefore, only the the "resource under test" should have planned changes.

https://github.com/hashicorp/terraform-plugin-testing/blob/88a9bc9b2c440b6e0f632721e0ca69f583a13c28/helper/resource/testing_new_import_state.go#L205-L229

This PR also makes a couple changes to turn off anything that expects plannable import to produce a new state. A "plan-only" plannable import has sufficed so far.

* Error on use of `ImportStatePersist` with plannable import

https://github.com/hashicorp/terraform-plugin-testing/blob/88a9bc9b2c440b6e0f632721e0ca69f583a13c28/helper/resource/testing_new_import_state.go#L430-L431

* Error on use of `ImportStateVerify` with plannable import
  * I was puzzled that `ImportStateVerify` was error-free for plannable import tests, despite the plan-only mechanics of these tests (no apply, no new state). I found that these tests trivially satisfied `ImportStateVerify` -- a loop over an empty `newResources` slice does nothing 😃

https://github.com/hashicorp/terraform-plugin-testing/blob/88a9bc9b2c440b6e0f632721e0ca69f583a13c28/helper/resource/testing_new_import_state.go#L433-L434

cc: @austinvalle who paired with me on 💭 and design
cc: @gdavison who raised the zone-domain example


helper/resource/testing_new_import_state.go
```diff
@@ -189,36 +213,40 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
-							if err != nil {
-								return fmt.Errorf("retrieving formatted plan output: %w", err)
-							}
-
-							return fmt.Errorf("importing resource %s: expected a no-op resource action, got %q action with plan \nstdout:\n\n%s", rc.Address, action, stdout)
-						}
-					}
+		// Verify reasonable things about the plan
+		var resourceChangeUnderTest *tfjson.ResourceChange
+		if kind.plannable() {
```
bbasata, line 220: oh, we're already plannable (line 166) :) this conditional is always true.

helper/resource/testing_new_import_state.go
```diff
@@ -279,7 +317,7 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 			identifierAttribute = "id"
 		}
 
-		for _, r := range newResources {
+		for _, r := range newResources { // This is a problem when newResources is empty!
```
bbasata, line 320: The `if !found` logic prevents false-negatives when `newResources` is empty.

I was puzzled that `ImportStateVerify` was 🍏 for some plannable import tests despite there being no apply, thus no new state.

This explains why they were trivially satisfying `ImportStateVerify`.

I'll add a guard to error early: if plannable, then ImportStateVerify is not supported.

bbasata (commented):
notes to self

helper/resource/importstate/import_block_for_resource_with_a_dependency_test.go
```diff
@@ -0,0 +1,49 @@
...
+resource "examplecloud_zone" "zone" {
+	name = "example.net"
+}
+
+resource "examplecloud_zone_record" "record" {
+	zone_id = examplecloud_zone.zone.id
+	name    = "www"
+}
+`
+	r.UnitTest(t, r.TestCase{
```
bbasata, line 29: TODO: require terraform 1.0+

.changes/unreleased/BREAKING CHANGES-20250404-170656.yaml
austinvalle, line 1: This is less of a nit and more of a discussion our team should probably have at some point.

I'm not really sure how to do the changelogs for alpha/beta releases. I've currently been doing just one or two `NOTE` changelogs mentioning the general features and not specific deltas between releases, but I can see why you'd want to be more explicit in-case anyone was trying the features out 🤔

helper/resource/importstate/import_block_for_resource_with_a_dependency_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	r "github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testprovider"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/providerserver"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestImportBlockForResourceWithADependency(t *testing.T) {
```
austinvalle, line 16: Nice test! 🎉

helper/resource/testing_new_import_state.go
```diff
@@ -145,21 +140,36 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
 
 	err = importWd.SetConfig(ctx, testStepConfig, step.ConfigVariables)
 	if err != nil {
 		t.Fatalf("Error setting test config: %s", err)
 	}
 
-	if !step.ImportStatePersist {
+	if kind.plannable() {
+		if stepNumber > 1 {
+			err = importWd.CopyState(ctx, wd.StateFilePath())
```
austinvalle, line 157: > This shouldn't block the PR merging, just some thinking out loud

I'm trying to figure out if this is all we need to do 🤔. I think in most cases this should suffice, but wondering if we have other external providers included in the first step... are we going to need to copy more (lockfile, downloaded providers)? Maybe the entire directory?

Would the follow-up init handle this seamlessly anyways? I think the `state rm` command should be safe since it doesn't need to initialize anything.

Maybe an eventual new test case could prove this theory out using a utility provider.

helper/resource/testing_new_import_state.go
```diff
@@ -181,44 +190,43 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
-							return fmt.Errorf("importing resource %s: expected a no-op resource action, got %q action with plan \nstdout:\n\n%s", rc.Address, action, stdout)
-						}
-					}
-				}
+		// Verify reasonable things about the plan
+		var resourceChangeUnderTest *tfjson.ResourceChange
+
+		if len(plan.ResourceChanges) == 0 {
+			return fmt.Errorf("importing resource %s: expected a resource change, got no changes", resourceName)
+		}
```
austinvalle, line 207: It just now struck me reading this code that a resource change will appear even if it's a no-op 😆

nit - Maybe this error messaging could reflect what we actually mean, i.e, the resource was not found in the plan at all (change != planned to be updated)

austinvalle (approved):
Looks great! I left a couple nits/thoughts 🚀

.changes/unreleased/BREAKING CHANGES-20250404-170656.yaml
austinvalle: Also, I failed to call out the import test changes in the first changelog for alpha, so maybe a general note would be useful.

helper/resource/testing_new_import_state.go
bbasata: +1 for a test case with external providers. I can add this in a follow-up PR.

helper/resource/testing_new_import_state.go
bbasata: 👍 to a more specific error message. Technically, this is simply checking safety of traversing `plan.ResourceChanges`, so it's the means to checking what we want. So we can make the test feedback more about the "end" than the "means."

helper/resource/testing_new_import_state.go
bbasata: > Would the follow-up init handle this seamlessly anyways?

I ran a quick test of this. Observations:

- Contents of a work directory

```shell
$ find step_1/work1565457440 -type f
step_1/work1565457440/testdata/1/examplecloud_container.tf
step_1/work1565457440/testdata/2/examplecloud_container_import.tf
step_1/work1565457440/terraform.tfstate
step_1/work1565457440/tfplan
step_1/work1565457440/.terraform/providers/registry.terraform.io/hashicorp/random/3.7.1/darwin_arm64/terraform-provider-random_v3.7.1_x5
step_1/work1565457440/.terraform/providers/registry.terraform.io/hashicorp/random/3.7.1/darwin_arm64/LICENSE.txt
step_1/work1565457440/terraform_plugin_test.tf
step_1/work1565457440/.terraform.lock.hcl
```

- Yes, the follow-up init handles this seamlessly
- The follow-up external provider download is an optimization opportunity -- why download twice? 😃 
- What if `ImportState` used [Terraform workspaces](https://developer.hashicorp.com/terraform/cli/workspaces)?

PR #480, opened 2025-04-08
helper/resource: add `ImportBlockWithResourceIdentity` kind

bbasata:
To merge after #476. This pull request adds resource identity support to `ImportState` test steps. For example:
https://github.com/hashicorp/terraform-plugin-testing/blob/c2a5b46c1881524da4233d584039d12fffe7da24/helper/resource/importstate/import_block_with_resource_identity_test.go#L32-L48

In this example, the test author writes in-line config in Step 1. In Step 3 with `ImportBlockWithResourceIdentity`, the test appends an `import` block. The test constructs this `import` block using resource identity values from the Step 1 state:
https://github.com/hashicorp/terraform-plugin-testing/blob/c2a5b46c1881524da4233d584039d12fffe7da24/helper/resource/testing_new_import_state.go#L118-L126

The test generates a plan and asserts that the plan is a no-op import of a single resource:
https://github.com/hashicorp/terraform-plugin-testing/blob/c2a5b46c1881524da4233d584039d12fffe7da24/helper/resource/testing_new_import_state.go#L221-L227

Finally, the test compares the Step 3 identity values with the identity values in Step 1 state. 💭 I want to take another look to consider whether this is necessary or duplicative of the "no-op import" assertion. 💭

At the time of writing this PR, I observed that a generated plan does not include resource identity values. And I reasoned that it's safe to run `apply` by this point -- the test applies a no-op import plan to a "sandboxed" statefile. Cross-ref https://github.com/hashicorp/terraform-plugin-testing/pull/476 "Error on use of ImportStatePersist with plannable import"

https://github.com/hashicorp/terraform-plugin-testing/blob/c2a5b46c1881524da4233d584039d12fffe7da24/helper/resource/testing_new_import_state.go#L233-L250


.changes/unreleased/NOTES-20250414-095537.yaml
```diff
@@ -0,0 +1,5 @@
+kind: NOTES
+body: 'This alpha pre-release adds support for managed resource identity, which can be used with Terraform v1.12.0-beta1. Acceptance tests can use the `ImportBlockWithResourceIdentity` kind to exercise the import of a managed resource using its resource identity object values instead of using a string identifier.'
```
austinvalle, line 2: I think we can just go ahead and release a beta at this point, aligning with core (indicating our general confidence in the approach)

```suggestion
body: 'This beta pre-release adds support for managed resource identity, which can be used with Terraform v1.12.0-beta1. Acceptance tests can use the `ImportBlockWithResourceIdentity` kind to exercise the import of a managed resource using its resource identity object values instead of using a string identifier.'
```

helper/resource/importstate/import_block_with_resource_identity_test.go
```diff
@@ -0,0 +1,87 @@
...
+			{
+				Config: `
+				resource "examplecloud_container" "test" {
+					location = "westeurope"
+					name     = "somevalue"
+				}`,
+			},
+			{
+				RefreshState: true,
+			},
```
austinvalle, line 43: Is this actually required? I'd expect just step 1 to suffice 🤔

helper/resource/importstate/import_block_with_resource_identity_test.go
```diff
@@ -0,0 +1,87 @@
...
```
austinvalle, line 78: Same question as above

helper/resource/testing.go
```diff
@@ -467,10 +467,16 @@ const (
 	ImportBlockWithResourceIdentity
 )
 
+// plannable reports whether this kind indicates the use of plannable import blocks
 func (kind ImportStateKind) plannable() bool {
 	return kind == ImportBlockWithID || kind == ImportBlockWithResourceIdentity
 }
 
+// resourceIdentity reports whether this this kind indicates the use of resource identity in import blocks
```
austinvalle, line 475:
```suggestion
// resourceIdentity reports whether this kind indicates the use of resource identity in import blocks
```

helper/resource/plugin.go
```diff
@@ -113,6 +114,58 @@ type providerFactories struct {
 	protov6 protov6ProviderFactories
 }
 
+func runProviderCommandApply(ctx context.Context, t testing.T, wd *plugintest.WorkingDir, factories *providerFactories) error {
```
bbasata, line 117: This is a light refactoring: these convenience functions remove incidental details from the main code flow.

helper/resource/testing_new_import_state.go
```diff
@@ -410,16 +434,45 @@ func appendImportBlock(config string, resourceName string, importID string) stri
 		resourceName, importID)
 }
 
+func appendImportBlockWithIdentity(config string, resourceName string, identityValues map[string]any) string {
```
austinvalle, line 437: Since we are "generating config" here, should we add some tests (like in `helper/resource/importstate/import_block_with_resource_identity_test.go`) for the differing types that we might encounter? We have string attribute encoding covered, so maybe bool, int/float, list of int/float or strings, etc?

helper/resource/testing_new_import_state.go
```diff
@@ -231,16 +231,40 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
 			return err
 		}
-	} else {
-		err = runProviderCommand(ctx, t, func() error {
-			return importWd.Import(ctx, resourceName, importId)
-		}, importWd, providers)
-		if err != nil {
-			return err
+
+		{
```
austinvalle, line 235: Was this extra scope block meant to stay?

helper/resource/testing_new_import_state.go
```diff
@@ -231,16 +231,40 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+				newStateJSON, err := runProviderCommandGetStateJSON(ctx, t, wd, providers)
+				if err != nil {
+					return fmt.Errorf("getting state after applying plan with import config: %s", err)
+				}
+
+				newIdentityValues := identityValuesFromState(newStateJSON, resourceName)
+				if !cmp.Equal(priorIdentityValues, newIdentityValues) {
+					return fmt.Errorf("importing resource %s: expected identity values %v, got %v", resourceName, priorIdentityValues, newIdentityValues)
+				}
+			}
```
austinvalle, line 251: I feel like I'm more leaning towards not applying and just waiting for TF core to give us the data during plan for comparison (it sounds like it should be included)

At least how it is ATM, core is validating that the data doesn't change during apply, and if they remove that logic, the SDKs should probably add it.

austinvalle: Mostly because we can't fully trust providers to apply what they plan (since the testing framework also supports SDKv2 😆 ), so if we can avoid it, I think we should

helper/resource/plugin.go
austinvalle: Love it!

helper/resource/testing_new_import_state.go
```diff
@@ -410,16 +434,45 @@ func appendImportBlock(config string, resourceName string, importID string) stri
...
 	versionUnderTest := *helper.TerraformVersion()
+	resourceIdentityMinimumVersion := version.Must(version.NewVersion("1.12.0-beta1"))
 
 	// Instead of calling [t.Fatal], we return an error. This package's unit tests can use [TestStep.ExpectError] to match
 	// on the error message. An alternative, [plugintest.TestExpectTFatal], does not have access to logged error messages,
 	// so it is open to false positives on this complex code path.
+	//
+	// Multiple cases may match, so check the most specific cases first
 	switch {
+	case kind.resourceIdentity() && versionUnderTest.LessThan(resourceIdentityMinimumVersion):
```
austinvalle, line 469: In the interest of this being out-of-date sometime next week when they release `v1.12.0-beta2` 😆, should we just compare the core version directly here? Then we can get out of the game of verifying that they are using a specific pre-release version:

We have some precedence for doing this, specifically for the `tfversion` checks: https://github.com/hashicorp/terraform-plugin-testing/blob/bfd5be787814353a166ee6c5723d327b4839ccc3/tfversion/skip_below.go#L42-L53

austinvalle (commented):
The code is looking great! I do have some specific questions and suggestions around the version check

helper/resource/importstate/import_block_with_resource_identity_test.go
bbasata: Yes, Step 1 is sufficient. I was simply happy that this worked 😃  

For clarity of the test's intent, I'm +1 on removing this. I'm starting to update the testing patterns docs, and I think that's the better place for a fleshed-out example like this one.

helper/resource/importstate/import_block_with_resource_identity_test.go
bbasata: Ditto 😃

helper/resource/testing_new_import_state.go
bbasata: I like it 💯

helper/resource/testing_new_import_state.go
bbasata: 👍 I was looking for a simple way to compare the core versions. I'll try to borrow this.

helper/resource/testing_new_import_state.go
bbasata: Sounds good to me. Since we have a working implementation in this PR, it should be no trouble to switch to planned values & expect tests to remain green.

helper/resource/testing_new_import_state.go
bbasata: https://github.com/hashicorp/terraform-plugin-testing/pull/480/commits/9c6616f743b2715f3bd28abc284c860f2f5cc9c1

helper/resource/importstate/import_block_with_resource_identity_test.go
bbasata: 33e8faf8

helper/resource/testing_new_import_state.go
bbasata: For the moment, yes, but not for long. Like the TODO comment that comes after this logic branch, there's a bit of tidying we can do to extract chunks of work from the long function.

austinvalle: Fair enough, just checking 😆

helper/resource/testing_new_import_state.go
bbasata: 09e574a3

helper/resource/testing_new_import_state.go
```diff
@@ -410,16 +435,64 @@ func appendImportBlock(config string, resourceName string, importID string) stri
...
+
+	for k, v := range identityValues {
+		switch v.(type) {
+		case bool:
+			configBuilder += fmt.Sprintf(`		%q = %t`+"\n", k, v)
+
+		case []any:
+			var quotedV []string
+			for _, v := range v.([]any) {
+				quotedV = append(quotedV, fmt.Sprintf(`%q`, v))
```
bbasata, line 454: 💭 also need to handle lists of numbers and lists of booleans

bbasata: or not? I added list of numbers and lists of booleans to the test resource schema and ... still green 🟢.

helper/resource/testing_new_import_state.go
bbasata: > 09e574a

and subsequent commits 😅

helper/resource/importstate/examplecloud_test.go
```diff
@@ -188,3 +231,251 @@ func examplecloudZoneRecord() testprovider.Resource {
...
+					"active":  tftypes.NewValue(tftypes.Bool, true),
+					"tags": tftypes.NewValue(
+						tftypes.List{ElementType: tftypes.String}, []tftypes.Value{
+							tftypes.NewValue(tftypes.String, "storage"),
+							tftypes.NewValue(tftypes.String, "fast")}),
+					"magic_numbers": tftypes.NewValue(
+						tftypes.List{ElementType: tftypes.Number}, []tftypes.Value{
+							tftypes.NewValue(tftypes.Number, 5),
+							tftypes.NewValue(tftypes.Number, 2)}),
+					"beep_boop": tftypes.NewValue(
```
bbasata, line 422: Really really reaching for an list of booleans example 😃

helper/resource/testing_new_import_state.go
```diff
@@ -407,16 +432,64 @@ func appendImportBlock(config string, resourceName string, importID string) stri
...
+			configBuilder += fmt.Sprintf(`		%q = %t`+"\n", k, v)
+
+		case []any:
+			var quotedV []string
+			for _, v := range v {
+				quotedV = append(quotedV, fmt.Sprintf(`%q`, v))
+			}
+			configBuilder += fmt.Sprintf(`		%q = [%s]`+"\n", k, strings.Join(quotedV, ", "))
+
+		case json.Number:
```
bbasata, line 455: TIL: `type json.Number string`

helper/resource/importstate/import_block_as_first_step_test.go
```diff
@@ -41,7 +41,13 @@ func TestImportBlock_AsFirstStep(t *testing.T) {
 				Config: `resource "examplecloud_container" "test" {
 					name = "somevalue"
 					location = "westeurope"
-				}`,
+				}
+
+				import {
```
bbasata, line 46: This test change is a consequence of #484.

Curiously, the tests did not run in https://github.com/hashicorp/terraform-plugin-testing/pull/484 💭

helper/resource/importstate/import_block_with_id_test.go
```diff
@@ -82,7 +82,15 @@ func TestImportBlock_WithID_ExpectError(t *testing.T) {
 				resource "examplecloud_container" "test" {
 					location = "eastus"
 					name     = "somevalue"
-				}`,
+				}
+
+				import {
```
bbasata, line 87: This test change is a consequence of #484.

Curiously, the tests did not run in https://github.com/hashicorp/terraform-plugin-testing/pull/484 💭

helper/resource/importstate/import_block_with_id_test.go
```diff
@@ -288,6 +290,12 @@ func TestImportBlock_WithID_WithBlankOptionalAttribute_GeneratesCorrectPlan(t *t
 
 				resource "examplecloud_container" "test" {
 					name = "somename"
+				}
+
+				import {
```
bbasata, line 295: This test change is a consequence of #484.

Curiously, the tests did not run in https://github.com/hashicorp/terraform-plugin-testing/pull/484 💭

helper/resource/testing_new_import_state.go
```diff
@@ -109,12 +108,20 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
 	if testStepConfig == nil && step.ConfigFile == nil && step.ConfigDirectory == nil {
 		logging.HelperResourceTrace(ctx, "Using prior TestStep Config for import")
 		importConfig := cfgRaw
 
-		if kind.plannable() {
+		if kind.plannable() && kind.resourceIdentity() {
+			priorIdentityValues = identityValuesFromState(stateJSON, resourceName)
+			if len(priorIdentityValues) == 0 {
+				return fmt.Errorf("importing resource %s: expected prior state to have resource identity values, got none", resourceName)
+			}
```
austinvalle, line 122: I believe this logic would need to be moved to another location outside the `if testStepConfig == nil && step.ConfigFile == nil && step.ConfigDirectory == nil` condition.

If you're providing a `step.ConfigFile`, it won't grab the prior identity values, then when we go to compare we'll end up comparing an empty map to whatever the identity actually is.

helper/resource/testing_new_import_state.go
```diff
@@ -228,16 +229,40 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+			if kind.resourceIdentity() {
+				err := runProviderCommandApply(ctx, t, wd, providers)
+				if err != nil {
+					return fmt.Errorf("applying plan with import config: %s", err)
+				}
+
+				newStateJSON, err := runProviderCommandGetStateJSON(ctx, t, wd, providers)
+				if err != nil {
+					return fmt.Errorf("getting state after applying plan with import config: %s", err)
+				}
```
austinvalle, line 243: I was testing this out and I believe these commands should be run against `importWd`?

bbasata: Yes, 💯. I've pushed a new commit to correct this.

helper/resource/testing_new_import_state.go
bbasata: Spot on. Fixed, with a new test.

helper/resource/testing_new_import_state.go
austinvalle: We talked about this offline and we're good with keeping apply with the caveat that we'll remove it before full release

.changes/unreleased/NOTES-20250414-095537.yaml
```diff
@@ -0,0 +1,5 @@
+kind: NOTES
+body: 'This beta pre-release adds support for managed resource identity, which can be used with Terraform v1.12.0-beta1. Acceptance tests can use the `ImportBlockWithResourceIdentity` kind to exercise the import of a managed resource using its resource identity object values instead of using a string identifier.'
```
austinvalle, line 2: 😆 

```suggestion
body: 'This beta pre-release adds support for managed resource identity, which can be used with Terraform v1.12.0-beta2. Acceptance tests can use the `ImportBlockWithResourceIdentity` kind to exercise the import of a managed resource using its resource identity object values instead of using a string identifier.'
```

helper/resource/importstate/examplecloud_test.go
austinvalle: Love it 😆

helper/resource/testing_new_import_state.go
```diff
@@ -134,31 +144,31 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 		}
 	}
 
-	var importWd *plugintest.WorkingDir
+	var workingDir *plugintest.WorkingDir
 
 	// Use the same working directory to persist the state from import
 	if importStatePersist {
-		importWd = wd
+		workingDir = testCaseWorkingDir
```
austinvalle, line 151: > Personally I don't think this is a big problem, but a little hidden trap is here I just thought about 🤔 

Since we are eventually planning on not applying the plannable import kinds (well the plannable block with ID already has this behavior), this persisted state would have a removed resource from it 😄.

If we remove the apply completely for import by identity, maybe we should just reject any combinations with `ImportStatePersist` (we chatted about this before i think? maybe it's already that way?)

austinvalle (approved):
LGTM, great work on this @bbasata!

helper/resource/testing_new_import_state.go
bbasata: Yes, we have it covered 😃 

https://github.com/hashicorp/terraform-plugin-testing/blob/88a9bc9b2c440b6e0f632721e0ca69f583a13c28/helper/resource/testing_new_import_state.go#L430-L431

austinvalle (approved):
🚀

PR #484, opened 2025-04-15
helper/resource: refine config-building for plannable import tests

bbasata:
This change allows explict inline Terraform configuration in an `ImportState` test step. In a newly-added test, we have a scenario where it's useful to the test author to write their own import block, without resorting to `ConfigFile` or `ConfigDirectory`.

This change also adds a test to cover the use of external providers with plannable import tests.

### DIFF
--- a/helper/resource/importstate/import_block_with_id_test.go
+++ b/helper/resource/importstate/import_block_with_id_test.go
@@ -8,7 +8,10 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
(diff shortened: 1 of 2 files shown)


helper/resource/importstate/import_block_with_id_test.go
```diff
@@ -376,3 +379,50 @@ func TestImportBlock_WithID_WithBlankComputedAttribute_GeneratesCorrectPlan(t *t
...
+			},
+		},
+		Steps: []r.TestStep{
+			{
+				Config: config,
+			},
+			{
+				ImportState:     true,
+				ImportStateKind: r.ImportBlockWithID,
+				Config:          configWithImportBlock,
```
austinvalle, line 415: Just to confirm my understanding, provider developers wouldn't need to use a config in this scenario to do this (if the ID needed to be explicitly defined), they could just use `ImportStateId`:

```go
			{
				ImportState:     true,
				ImportStateKind: r.ImportBlockWithID,
				ImportStateId:   "It was a dark and stormy night.",
				ResourceName:    "random_string.mystery_message",
				ImportPlanChecks: r.ImportPlanChecks{
					PreApply: []plancheck.PlanCheck{
						plancheck.ExpectKnownValue(
							"random_string.mystery_message",
							tfjsonpath.New("result"),
							knownvalue.StringExact("It was a dark and stormy night.")),
					},
				},
```

austinvalle (approved):
LGTM 🚀

helper/resource/importstate/import_block_with_id_test.go
bbasata: > `ImportStateId`

Nice catch. We could come up with a different example here so that it better models a situation that calls for an inline config.

PR #488, opened 2025-04-18
Remove test dependency on checkpoint api response

bbasata:


helper/resource/teststep_providers_test.go
```diff
@@ -3361,36 +3361,31 @@ func TestTest_TestStep_ProviderFactories_Import_External_With_Data_Source(t *tes
 	Test(t, TestCase{
 		ExternalProviders: map[string]ExternalProvider{
 			"http": {
-				Source: "registry.terraform.io/hashicorp/http",
+				Source: "registry.terraform.io/hashicorp/null",
```
austinvalle, line 3364: Looks like the key in this map just needs to be updated to `null` 👍🏻

austinvalle (approved):

PR #493, opened 2025-04-24
experimental: faster tfversion tests

bbasata:
For fun, not to be merged.

The `tfversion` tests are fairly unique in that they specify `"TF_ACC_TERRAFORM_VERSION"`. As a consequence, each of these tests downloads Terraform into its own temporary directory. This PR asks: what if they shared one temporary directory?

* Find Terraform binaries that we've already downloaded instead of repeatedly downloading them
* Read a `TF_LOG_INSTALLER` environment variable to log hc-install activity. Works best with a branch of hc-install that prepends log levels to logged messages.
* What if ... provider installation can also be de-duplicated?

Before:
```
$ time go test ./tfversion -count=1
ok  	github.com/hashicorp/terraform-plugin-testing/tfversion	223.588s
go test ./tfversion -count=1  45.66s user 29.61s system 33% cpu 3:44.32 total
```

After:
```
$ time go test ./tfversion -count=1
ok  	github.com/hashicorp/terraform-plugin-testing/tfversion	58.465s
go test ./tfversion -count=1  22.14s user 12.61s system 58% cpu 58.982 total
```

### DIFF
--- a/internal/plugintest/config.go
+++ b/internal/plugintest/config.go
@@ -7,8 +7,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-version"
 	install "github.com/hashicorp/hc-install"
 	"github.com/hashicorp/hc-install/checkpoint"
@@ -16,6 +18,7 @@ import (
 	"github.com/hashicorp/hc-install/product"
 	"github.com/hashicorp/hc-install/releases"
 	"github.com/hashicorp/hc-install/src"
+	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
 
 	"github.com/hashicorp/terraform-plugin-testing/internal/logging"
 )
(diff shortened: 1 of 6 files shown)


internal/plugintest/config.go
```diff
@@ -60,11 +66,31 @@ func DiscoverConfig(ctx context.Context, sourceDir string) (*Config, error) {
...
+			Product:    product.Terraform,
+			Version:    tfVersion,
+		}
+		releasesSource := &releases.ExactVersion{
 			InstallDir: tfDir,
 			Product:    product.Terraform,
 			Version:    tfVersion,
-		})
+		}
+		sources = append(sources, findSource, releasesSource)
```
bbasata, line 92: This is the part that makes this faster. Instead of a "releases source," use a "find source" and a "releases source."

internal/plugintest/config.go
```diff
@@ -60,11 +66,31 @@ func DiscoverConfig(ctx context.Context, sourceDir string) (*Config, error) {
...
 
-		sources = append(sources, &releases.ExactVersion{
+		if tempDir == "" {
+			tempDir = os.TempDir()
+		}
+		pathParts := []string{
+			strings.TrimRight(tempDir, string(os.PathSeparator)),
+			"plugintest-terraform",
+			strconv.Itoa(os.Getpid()),
+			tfVersion.String(),
```
bbasata, line 76: One directory per-version, per-process.

Issue #383, opened 2024-10-31
Enhancement: Allow composition of config directories


jar-b:
### terraform-plugin-testing version

```
github.com/hashicorp/terraform-plugin-testing v1.10.0
```

### Use cases

In the Terraform AWS provider, we frequently have acceptance tests which share common setup infrastructure across tests. To avoid repeating the setup, we have a helper function ([`acctest.ConfigCompose`](https://github.com/hashicorp/terraform-provider-aws/blob/720fda7f1bc49358a160633cd5f4b3b9d6ffa53a/internal/acctest/acctest.go#L1902-L1911)) which composes a variable number of configuration strings into a single string which is then passed to the `TestStep.Config` argument. The setup is roughly:

```go
func TestAccSomeResource_basic(t *testing.T) {
	resource.ParallelTest(t, resource.TestCase{
		// Other setup
		Steps: []resource.TestStep{
			{
				Config: testAccSomeResource_basic(rName),
				// Other checks
			},

		},
	})
}


func testAccSomeResourceConfigBase(rName string) string {
	return fmt.Sprintf(`
data "aws_other_resource" "test" {
  name = %[1]q
}
`, rName)
}

func testAccSomeResourceConfig_basic(rName string) string {
	return acctest.ConfigCompose(
		testAccSomeResourceConfigBase(rName),
		fmt.Sprintf(`
resource "aws_some_resource" "test" {
  name      = %[1]q
  other_arn = aws_other_resource.test.arn
}
`, rName))
}

// Composition is repeated for additional test configurations
```

As we are investigating adoption of the new `ConfigDirectory` test step option, we're finding that migrating from the pattern above requires copying the "setup" configuration into the `testdata` subdirectory for each test. If we need to make a change to the setup, we now have to apply that change several times (each `testdata` subdirectory), rather than in a single location. 

### Attempted solutions

- Copying "setup" configuration into each `testdata` directory

### Proposal

A helper function to compose directories would allow us to replicate this pattern while gaining the benefits of `ConfigDirectory`/`ConfigFile` over HCL strings embedded within our Go source code. 

```go
func TestAccSomeResource_basic(t *testing.T) {
	resource.ParallelTest(t, resource.TestCase{
		// Other setup
		Steps: []resource.TestStep{
			{
				ConfigDirecotory: config.ComposeDirectories(,
				        config.StaticDirectory("testdata/setup"),
				        config.TestNameDirectory(),
				),
				// Other checks
			},

		},
[...]

PR #494, opened 2025-04-28
plannable import: support `step.ConfigExact = [true|false]` for all config sources

bbasata:
This PR extends `ImportState` auto-generated import blocks to work with `step.ConfigFile` and `step.ConfigDirectory`. It adds a `step.ConfigExact` flag to opt out.

cc: @gdavison


bbasata:
cc: @austinvalle, I'd like to get your input on how this feels directionally, so we can decide whether to ship it in a release candidate.

internal/teststep/config.go
```diff
@@ -45,6 +45,7 @@ type Config interface {
 	HasProviderBlock(context.Context) (bool, error)
 	HasTerraformBlock(context.Context) (bool, error)
 	Write(context.Context, string) error
+	Append(context.Context, string) Config
```
bbasata, line 48: This feels simpler than `teststep.Configuration(teststep.PrepareConfigurationRequest({...}))`.

The thought is: given a `Config` value, produce a new `Config` value that includes the appended generated config.

And then each concrete type implements how to write its "base" config and the appended generated config.

For a string: concatenate.

For a file: append to the file.

For a directory: add a file to the directory.

austinvalle: Very clean, like it 👍🏻, and it looks like we just hold that in a buffer until we read the data which is good.

austinvalle, line 48: I can't remember, are we guaranteed to always be writing to a temporary test directory / test file? I think we copy them to a temporary directory, but I can't remember the details.

Just want to make sure our eventual changes aren't going to be rewriting/appending to the actual source test configuration files 👍🏻

austinvalle (commented):
TLDR, big fan of the approach/feedback if it works well for provider developers and think we should expand on it 👍🏻

internal/teststep/config.go
bbasata: 💯

helper/resource/testing.go
```diff
@@ -666,6 +666,10 @@ type TestStep struct {
 
 	ImportStateKind ImportStateKind
 
+	// ImportStateReadOnlyConfig indicates that the test framework should not
+	// modify ConfigFile or ConfigDirectory inputs to the test step.
+	ImportStateReadOnlyConfig bool
```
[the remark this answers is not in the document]
bbasata: > should ImportStateReadOnlyConfig also apply to the Config string?

This feels right to me.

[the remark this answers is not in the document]
bbasata: > If we can't come up with a good name I can also live with just the pkg documentation being explicit. 

👍 This PR plays with different words for kinda the same thing, so we have options. "ReadOnly," "Append", "Generated." I'm leaning away from "ReadOnly" ...

> The `TestStep` struct is already pretty far gone

Next time ... 😃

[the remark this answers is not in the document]
bbasata: Proposed in the latest commit: `step.ConfigExact`.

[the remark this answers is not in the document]
bbasata: 💭 an alternative that does not need a new struct field:

Screenshot 2025-05-02 at 12 12 38 PM

internal/teststep/config.go
bbasata: Yes, double-checked now.

File:

https://github.com/hashicorp/terraform-plugin-testing/blob/4a1b9f2cd2c7126226c7d3a54ef30dcde2c53060/internal/teststep/file.go#L87

Directory:

https://github.com/hashicorp/terraform-plugin-testing/blob/4a1b9f2cd2c7126226c7d3a54ef30dcde2c53060/internal/teststep/directory.go#L87

helper/resource/testing_new_import_state.go
```diff
@@ -116,23 +116,40 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+		// TODO: extract / DRY
+
+		if kind.plannable() && kind.resourceIdentity() {
+			testStepConfig = appendImportBlockWithIdentity(testStepConfig, resourceName, priorIdentityValues)
+		} else if kind.plannable() {
+			testStepConfig = appendImportBlock(testStepConfig, resourceName, importId)
+		}
+
+	case step.ConfigFile != nil:
+		// TODO: ship it
```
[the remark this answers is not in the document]
bbasata: Done!

bbasata:
> TLDR, big fan of the approach/feedback if it works well for provider developers and think we should expand on it 👍🏻

I re-discovered https://github.com/hashicorp/terraform-plugin-testing/issues/383 just now.

gdavison:
Yes, #383 will be very useful for us. In the string-based `Config` flow, we append, sometimes several levels deep, common configuration. Being able to add files or merge directories would simplify the use of `ConfigDirecotory`

helper/resource/testing.go
```diff
@@ -576,6 +576,16 @@ type TestStep struct {
...
 
+	// ConfigExact indicates that the test framework should use the exact
+	// content of the Config, ConfigFile, or ConfigDirectory inputs and should
+	// not modify it at test run time.
+	//
+	// The default is false. At test run time, the test framework will generate
+	// specific kinds of configuration, such as import blocks, and append them
+	// to the given Config, ConfigFile, or ConfigDirectory inputs. Using this
+	// default improves test readability and removes duplication of setup.
+	ConfigExact bool
```
austinvalle, line 587: The only qualm I have with this naming is that currently, the default for `Import` mode is that this is `false`... however if you apply this to `Config` mode, the default is technically true for `ConfigFile/Directory` and false for `Config` right?

We wouldn't be able to, in the future, make modifications based off this flag in `Config` mode because it would be a breaking change 🤔.

That's all from my memory of how we treat `Config` mode, so might need to be verified 😆, I just want to make sure we don't put ourselves in a corner

helper/resource/testing_new_import_state.go
```diff
@@ -105,34 +105,36 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
...
+		Raw:                   inlineConfig,
 		TestStepConfigRequest: testStepConfigRequest,
 	}.Exec())
 
-	if testStepConfig == nil {
-		logging.HelperResourceTrace(ctx, "Using prior TestStep Config for import")
-		importConfig := cfgRaw
+	switch {
+	case step.ConfigExact:
+		break
```
austinvalle, line 127: Does this fallthrough by default if there is no body? My yearly reminder of the `fallthrough` Go directive 😆 

I like the explicitness regardless

internal/teststep/config.go
```diff
@@ -187,12 +188,28 @@ func copyFile(path string, dstPath string) error {
...
+		return "", err
 	}
 
 	defer dstF.Close()
 
 	if _, err := io.Copy(dstF, srcF); err != nil {
+		return "", err
+	}
+
+	return dstPath, nil
```
austinvalle, line 200: Is there a reason that we're just re-returning a string that was passed in without modifying it?

internal/teststep/file.go
```diff
@@ -84,11 +85,24 @@ func (c configurationFile) Write(ctx context.Context, dest string) error {
...
 
-	err := copyFile(configFile, dest)
-
+	destPath, err := copyFile(configFile, dest)
 	if err != nil {
 		return err
 	}
 
+	if len(c.appendedConfig) > 0 {
+		err := appendToFile(destPath, c.appendedConfig)
```
austinvalle, line 94: Wouldn't `destPath` always be equal to `dest` unless there is an error? (in which case we should have returned prior to this I think)

internal/teststep/directory.go
```diff
@@ -5,14 +5,20 @@ package teststep
...
 	"context"
+	"fmt"
+	"hash/crc32"
 	"os"
 	"path/filepath"
 )
 
 var _ Config = configurationDirectory{}
 
+// not threadsafe
```
austinvalle, line 16: Does it matter that this isn't thread safe? Wondering if this comment is an omen or just left over 😆

helper/resource/testing.go
austinvalle: Put differently, is there a reason we shouldn't just name this `ImportStateConfigExact` (besides how icky the name is 😆)

austinvalle (commented):
I have no major qualms with the approach, just a thought on naming

helper/resource/testing_new_import_state.go
bbasata: Nice callout 😃 `break` is indeed optional here. I likely wrote `break` explicitly due to habit. To your point, yes, explicitness does feel more intention-revealing here & kinder to our future selves.

I've confirmed that the tests pass without the `break` statement, and they fail when I replace it with `fallthrough`.

internal/teststep/directory.go
bbasata: Right, I think it doesn't matter much & is an artifact of the design process.

We can remove the comment ✅

helper/resource/testing.go
bbasata: > We wouldn't be able to, in the future, make modifications based off this flag in `Config` mode because it would be a breaking change 🤔.

Yes, I like this angle... design consequences 😃

Option: keep this as a global non-prefixed option, and explicitly update `Config` and `Refresh` mode to error when `true`. Then we have the future option to accept that flag in either mode without breaking tests. 

> Put differently, is there a reason we shouldn't just name this `ImportStateConfigExact` (besides how icky the name is 😆)

Also an option. I'm willing to go this route & encode the fact that we simply don't know how general this concept should be.

I'll go with that and 💭 on the naming for a moment.

internal/teststep/config.go
bbasata: `dstPath` is conditionally modified on line 185. Adding a return value was a compromise to allow us to pass the exact same path in `appendToFile`.

There's a refactoring opportunity there; since it's all `internal`, I'm ok with shipping it and revisiting the next time we touch this.

internal/teststep/file.go
bbasata: Does https://github.com/hashicorp/terraform-plugin-testing/pull/494#discussion_r2077766538 address this?

internal/teststep/config.go
austinvalle: Oh whoops! I glossed over GitHub hiding more lines from me when reading 😆 
image

No need to revisit it 👍🏻

internal/teststep/file.go
austinvalle: Yes! Thanks for pointing that 👍🏻

helper/resource/testing.go
austinvalle: Sweet! I'm good with that 👍🏻

austinvalle (approved):
LGTM 🚀

Issue #269, opened 2024-01-08
Errors with `ImportStateVerify` test steps after `terraform-plugin-testing@v1.6.0` update


ewbankkit:
Ever since the `terraform-plugin-testing@v1.6.0` update a number of Terraform AWS Provider `ImportStateVerify` state verify test steps that previously passed have been failing.

For example:

```
=== RUN   TestAccAppMesh_serial/Route/grpcRouteEmptyMatch
    route_test.go:568: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
          map[string]string{
        +   "spec.0.grpc_route.0.action.0.weighted_target.0.port": "0",
          }
```

The pattern I have observed is that these tests cases are all to do with values of attributes inside nested `TypeSet`s that are themselves nested in other configuration blocks. The affected resource all use Terraform Plugin SDK v2.

For the example above (`internal/service/appmesh/route.go`):

```go
				"grpc_route": {
					Type:          schema.TypeList,
					Optional:      true,
					MinItems:      0,
					MaxItems:      1,
					ConflictsWith: []string{"spec.0.http2_route", "spec.0.http_route", "spec.0.tcp_route"},
					Elem: &schema.Resource{
						Schema: map[string]*schema.Schema{
							"action": {
								Type:     schema.TypeList,
								Required: true,
								MinItems: 1,
								MaxItems: 1,
								Elem: &schema.Resource{
									Schema: map[string]*schema.Schema{
										"weighted_target": {
											Type:     schema.TypeSet,
											Required: true,
											MinItems: 1,
											MaxItems: 10,
											Elem: &schema.Resource{
												Schema: map[string]*schema.Schema{
													"port": {
														Type:         schema.TypeInt,
														Optional:     true,
														ValidateFunc: validation.IsPortNumber,
													},
```

Acceptance test case (`testAccRoute_grpcRouteEmptyMatch`) configuration (`internal/service/appmesh/route_test.go`):

```hcl
resource "aws_appmesh_route" "test" {
  name                = "test"
  mesh_name           = aws_appmesh_mesh.test.id
  virtual_router_name = aws_appmesh_virtual_router.test.name

  spec {
    grpc_route {
      match {}

      action {
        weighted_target {
          virtual_node = aws_appmesh_virtual_node.test1.name
          weight       = 100
        }
      }
    }
  }
}
```

### terraform-plugin-testing version

```console
[...]

Issue #327, opened 2024-04-22
`ImportStateVerify` fails for `TypeSet` fields in terraform-plugin-testing but not terraform-plugin-sdk


trodge:
### terraform-plugin-testing version

```
github.com/hashicorp/terraform-plugin-testing v1.7.0
```

### Expected Behavior

Set fields should pass `ImportStateVerify` regardless of the order of elements as they do in terraform-plugin-sdk.

### Actual Behavior

`ImportStateVerify` fails with a different ordering of the same elements in a set field.

### Steps to Reproduce

Run any of the tests from [this list](https://github.com/GoogleCloudPlatform/magic-modules/pull/10339#issuecomment-2050693937) in the google provider after migrating to terraform-plugin-sdk.

### References

This issue was previously closed as being caused by an upstream library, but it was never explained what the upstream issue was or why the tests pass in terraform-plugin-sdk but not in terraform-plugin-testing

https://github.com/hashicorp/terraform-plugin-testing/issues/269

austinvalle:
Hey there @trodge 👋🏻, thanks for submitting the bug and sorry you're running into an issue here.

I don't have access to the CI logs posted to that linked PR, would you mind taking one/a couple of those error message logs and publishing them in this GitHub issue if possible? (I've been chatting with @SarahFrench internally, so you both may be already working on this)

As for #269, that bug report didn't come to a conclusion on the specific root cause, but was closed by the reporter as one of the likely causes of the behavior were related to the upstream `terraform-plugin-sdk/v2` library. There are a couple documented bugs in `terraform-plugin-sdk/v2` related to `TypeSet` that could be creating inconsistencies now displayed by `terraform-plugin-testing`. A lot of these bugs we can't fix for compatibility reasons, however if the problem is in `terraform-plugin-testing`, we might be able to.

That being said, it would still be useful to see the exact error messages and related SDKv2 schemas, so I can recreate the bug you're experiencing and then definitively describe what's happening and whether we can actually fix it.

trodge:
Here is the error message from one of the tests:
```
2024-04-11T20:50:05.607Z [ERROR] sdk.helper_resource: Error running import: test_working_directory=/tmp/plugintest2444090125
  error=
  | ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
  | 
  | \u00a0\u00a0map[string]string{
  | -\u00a0\t"access.0.iam_member":     "allUsers",
[...]

PR #496, opened 2025-05-02
helper/resource: compatibility refresh after config mode test step

bbasata:
Background: https://github.com/hashicorp/terraform-plugin-testing/issues/269

This change is intended to unlock an upgrade path from `v1.5.1` to latest `v1.13.1`.

When `RefreshAfterApply` is set to non-empty via a build flag, a `Config`-mode test step will invoke a refresh before successful completion. This is intended as a compatibility measure for test cases that have different -- but semantically-equal -- state representations in their test steps. When comparing two states, the testing framework is not aware of semantic equality or set equality, as that would rely on provider logic.

A `Config`-mode test step that is followed by a `Refresh`-mode test step will not incur an extra refresh. This is a backward-compatible way for a provider to eventually remove the compatibility flag – existing tests can be updated to explicitly refresh, incrementally.

cc: @BBBmau: I'd definitely like to get feedback from the provider team before shipping this. And it would be helpful to try this on a feature branch & see if it introduces any test failures.


austinvalle (commented):
Yeah, like we chatted about offline, if this can be a stop gap before https://github.com/hashicorp/terraform-plugin-testing/issues/327 is actually resolved, I'm on board 👍🏻 

Other related issues and the original change:
- https://github.com/hashicorp/terraform-plugin-testing/issues/269
- https://github.com/hashicorp/terraform-plugin-testing/pull/223

helper/resource/compatibility.go
```diff
@@ -0,0 +1,13 @@
...
+package resource
+
+// Deprecated. This is an undocumented compatibility flag. When
+// `RefreshAfterApply` is set to non-empty, a `Config`-mode test step will
+// invoke a refresh before successful completion. This is intended as a
+// compatibility measure for test cases that have different -- but
+// semantically-equal -- state representations in their test steps. When
+// comparing two states, the testing framework is not aware of semantic
+// equality or set equality, as that would rely on provider logic.
+var RefreshAfterApply string
```
bbasata, line 13: When I started playing with this, I deliberately chose a very long variable name as a feature flag. The thought is that you'd add this as one more `ldflags` value in a `testacc` Make target:
```
-X github.com/hashicorp/terraform-plugin-testing/helper/resource.DeprecatedCompatibilityFlagRefreshAfterApply=unlocked
```

The `ldflags` bit is a weakly-held opinion. I would be receptive to using a simpler name and a `TF_*` environment variable instead.

helper/resource/compatibility.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resource
+
+// Deprecated. This is an undocumented compatibility flag. When
```
bbasata, line 6: > `Deprecated`

This is about the flexibility to remove this flag some day. There is no timetable or specific urgency for that. If we get feedback that this flag is super effective & we do not come up with a better alternative in the future, then I have no desire to break it 😃

helper/resource/compatibility.go
austinvalle: Personally I'd lean toward an environment variable, just being consistent with existing configuration options: `TF_ACC_REFRESH_AFTER_APPLY` or something of the like 😆

austinvalle (commented):
Generally good with everything in this PR, since you prompted I would probably lean towards using an env variable, so if you're still keen to make that change I'll wait to approve

BBBmau (approved):
I've done some TC tests and it looks to be good on the google side for addressing the verison pin, any chance we can get a release cut sometime soon? We can finish up MM support for ResourceIdentity once it's done.

bbasata:
> I've done some TC tests and it looks to be good on the google side for addressing the verison pin, any chance we can get a release cut sometime soon? We can finish up MM support for ResourceIdentity once it's done.

@BBBmau: I'm sorry for my delay in replying. I've now updated this branch to use a new environment variable: `TF_ACC_REFRESH_AFTER_APPLY`.

Would you give this one more run with your provider's suite for confidence before we release?

BBBmau:
@bbasata did some 2 nightly runs to confirm that things are good with the environment variable! We can go ahead with the release when you're ready.

helper/resource/testing_new_config.go
```diff
@@ -443,5 +451,19 @@ func testStepNewConfig(ctx context.Context, t testing.T, c TestCase, wd *plugint
 		}
 	}
 
+	if refreshAfterApply && !step.Destroy && !step.PlanOnly {
+		if len(c.Steps) > stepIndex+1 {
+			// If the next step is a refresh, then we have no need to refresh here
+			if !c.Steps[stepIndex+1].RefreshState {
+				// Echo a searchable message to easily determine when this is no longer being used
+				fmt.Println(EnvTfAccRefreshAfterApply+":", "running apply -refresh-only -refresh=true")
```
austinvalle, line 459: nit: We should probably use the `logging` package for this with whatever log level you think is appropriate

austinvalle (approved):
Changes look good, one small nit.

We probably also want a changelog at some point before the release, if you'd prefer to do it in another PR that's cool too 👍🏻

austinvalle (approved):
LGTM, the tests are failing but it doesn't look related 👍🏻

austinvalle (approved):

PR #499, opened 2025-05-07
ImportState: Fix panic when identity attribute is `null`

austinvalle:
This PR introduces a change to import configuration building of identities to account for the scenario where an `OptionalForImport` attribute.

It's technically valid for any identity attribute to be `null`, regardless of the behavior field set, but if a provider developer wants to test import via identity for a different scenario, where the attribute is provided in import, they can always create a custom config.

--------

**[Test failure](https://github.com/hashicorp/terraform-plugin-testing/actions/runs/14893490445/job/41830997679) without the change:**
```bash
--- FAIL: TestImportBlock_WithResourceIdentity_NullAttribute (1.36s)
panic: unexpected type <nil> for identity value "value_we_dont_always_need" [recovered]
	panic: unexpected type <nil> for identity value "value_we_dont_always_need"
```


helper/resource/testing_new_import_state.go
```diff
@@ -458,8 +458,12 @@ func appendImportBlockWithIdentity(config string, resourceName string, identityV
 
 		case string:
 			configBuilder += fmt.Sprintf(`		%q = %q`+"\n", k, v)
-
 		default:
+			// It's valid for identity attributes to be null, we can just omit it from config
+			if v == nil {
+				break
+			}
```
austinvalle, line 465: Testing shows that this should be enough for checking `nil` in this exact use-case. Since the only producer of this data is `terraform-json`, feels like we won't need to also use reflection? 🤔

bbasata: Nice! Suggestion: it's safer (and/or less cognitive overhead) to check for `nil` before the `switch`, since `nil` values can have a type.

bbasata (approved):
:shipit:

PR #500, opened 2025-05-12
Prep changelogs for `v1.13.0` release

austinvalle:
Moved the changelogs out of the `upcoming-stable` folder, as well as refactored some of them (since we were originally going to release an RC, but didn't get around to it).


bbasata (approved):
:shipit:

PR #502, opened 2025-05-13
ResourceIdentity: Switch to using plan instead of applied state for asserting identity data

austinvalle:
The goal of the current assertion for identity imports is to ensure the prior saved identity is the same as the identity after import/refresh (i.e. the planned identity). This PR switches the previous logic from doing an apply, to using the `PlannedValues` of the no-op plan produced by a successful plannable import.

Luckily, `PlannedValues` also happens to be a `tfjson.StateValues`, similar to the `(tfjson.State).Values`, so there is a small refactor to do the comparison 👍🏻

### DIFF
--- a/helper/resource/plugin.go
+++ b/helper/resource/plugin.go
@@ -114,15 +114,6 @@ type providerFactories struct {
 	protov6 protov6ProviderFactories
 }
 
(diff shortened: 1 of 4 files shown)


helper/resource/plugin.go
austinvalle, line 1: These changes are just to satisfy the linter since they are no longer used 👍🏻

ansgarm (approved):

PR #503, opened 2025-05-13
statecheck: Add new resource identity / state comparison checks

austinvalle:
This PR adds two new state checks that asserts that an identity AND state value at the given path(s) are exactly equal. This was originally a suggestion from the Azure team that are looking to simplify their testing assertions.

cc @stephybun

### DIFF
--- a/statecheck/expect_identity_value_matches_state_example_test.go
+++ b/statecheck/expect_identity_value_matches_state_example_test.go
@@ -1,0 +1,38 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package statecheck_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func ExampleExpectIdentityValueMatchesState() {
+	// A typical test would accept *testing.T as a function parameter, for instance `func TestSomething(t *testing.T) { ... }`.
+	t := &testing.T{}
+	t.Parallel()
+
+	resource.Test(t, resource.TestCase{
+		// Resource identity support is only available in Terraform v1.12+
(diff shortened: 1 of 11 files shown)


statecheck/expect_identity_value_matches_state_example_test.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+		// Provider definition omitted. Assuming "test_resource":
+		//  - Has an identity schema with an "id" string attribute
+		//  - Has a resource schema with an "id" string attribute
+		Steps: []resource.TestStep{
+			{
+				Config: `resource "test_resource" "one" {}`,
+				ConfigStateChecks: []statecheck.StateCheck{
+					// The identity attribute and state attribute at "id" must match
+					statecheck.ExpectIdentityValueMatchesState("test_resource.one", tfjsonpath.New("id")),
```
stephybun, line 33: Would it be possible to extend this to allow providers to specify a different value for the identity attribute and state attribute?

austinvalle: By different value, do you mean providing two pathes one for identity and one for state?

```go
statecheck.ExpectIdentityValueMatchesStateAtPath("test_resource.one", tfjsonpath.New("identity_id"), tfjsonpath.New("state_id"))
```

We can definitely do that, just want to make sure I'm understanding what you're asking for

austinvalle: My other understanding of what you said would be providing a different value that both attributes must match, via a `knownvalue.Check`, which could be done in either variant:

```go
statecheck.ExpectIdentityValueMatchesState(
  "test_resource.one",
  tfjsonpath.New("id"),
  knownvalue.StringExact("id-123"),
)

statecheck.ExpectIdentityValueMatchesStateAtPath(
  "test_resource.one",
  tfjsonpath.New("identity_id"),
  tfjsonpath.New("state_id"),
  knownvalue.StringExact("id-123"),
)
```

austinvalle: Awesome, thanks for the feedback, I added that alternative check in 732e97ecdf87beea203aec3e4131a9a952455dcd ⚡

ansgarm (approved):

PR #518, opened 2025-06-02
importstate: Use `ExpectNonEmptyPlan` field to ignore no-op assertion for import block test kinds

austinvalle:
## Related Issue

N/A

## Description

This change adds the usage of `(TestStep).ExpectNonEmptyPlan` to ignore the assertion that all import block plans must produce a no-op. A no-op plan isn't possible for some resource types that cannot return the entire original value from the remote API after importing. The usage of `ImportPlanChecks` can be used to make more fine-grain assertions about the plan to ensure the plan produced during import is correct.

**Test failure prior to introducing the change:**
```bash
--- FAIL: TestImportBlock_WithID_ExpectNonEmptyPlan (0.75s)
    /Users/austin.valle/code/terraform-plugin-testing/helper/resource/importstate/import_block_with_id_test.go:105: Step 2/2 error running import: importing resource examplecloud_container.test: expected a no-op import operation, got ["update"] action with plan 
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
        Terraform will perform the following actions:
        
          # examplecloud_container.test will be updated in-place
          # (imported from "westeurope/somevalue")
          ~ resource "examplecloud_container" "test" {
              ~ id       = "westeurope/somevalue" -> (known after apply)
              ~ location = "westeurope" -> "eastus"
                name     = "somevalue"
            }
        
        Plan: 1 to import, 0 to add, 1 to change, 0 to destroy.
FAIL
FAIL	github.com/hashicorp/terraform-plugin-testing/helper/resource/importstate	1.407s
FAIL
```

## Rollback Plan

- [X] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Nope.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20250602-073221.yaml
+++ b/.changes/unreleased/BUG FIXES-20250602-073221.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'helper/resource: Updated `ImportBlockWith*` import state modes to use the `ExpectNonEmpty` field to allow non-empty import plans to pass successfully.'
+time: 2025-06-02T07:32:21.384247-04:00
+custom:
+    Issue: "518"


rainkwan (approved):
Looks reasonable, we can revisit the flagging question later if needed.

PR #519, opened 2025-06-02
importstate: fix reading configuration from previous step

jeremmfr:
## Related Issue

Fixes #516

## Description

Store the ConfigDirectory and ConfigFile from the current step, so they can be used by subsequent ImportState steps if not explicitly set

PR #522, opened 2025-06-06
importstate: Fixed bug where prior test config is not used for `ConfigFile` or `ConfigDirectory`

### LINKED ISSUE #516: 1.13.* : importstate doesn't use configuration from previous state with ConfigDirectory

```
github.com/hashicorp/terraform-plugin-testing v1.13.1
```

The issue is not present in v1.12.0

### Relevant provider source code

To reproduce the bug, I tested with https://github.com/hashicorp/terraform-provider-hashicups/tree/main/12-final after upgrading this module to v1.13.1, terraform-plugin-framework to v1.15.0 and terraform-plugin-go to v0.27.0

```go
func TestAccOrderResource(t *testing.T) {
		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
		Steps: []resource.TestStep{
			// Create and Read testing
			{
				ConfigDirectory: config.TestStepDirectory(),
			},
			// ImportState testing
			{
				ResourceName:      "hashicups_order.test",
				ImportState:       true,
				ImportStateVerify: true,
				// The last_updated attribute does not exist in the HashiCups
				// API, therefore there is no value for it during import.
				ImportStateVerifyIgnore: []string{"last_updated"},
			},
		},
	})
}
```

### Terraform Configuration Files

`12-final/internal/provider/testdata/TestAccOrderResource/1/main.tf`

```tf
resource "hashicups_order" "test" {
  items = [
    {
      coffee = {
        id = 1
      }
      quantity = 2
    },
  ]
}

```

### Expected Behavior

On the second step, the import doesn't catch an error

### Actual Behavior

On the second step, the module catch this error when using `ConfigDirectory: config.TestStepDirectory()` on the previous step before import : 

```bash
=== RUN   TestAccOrderResource
    order_resource_test.go:11: Step 2/2 error running import: exit status 1
        Error: resource address "hashicups_order.test" does not exist in the configuration.
        
        Before importing this resource, please create its configuration in the root module. For example:
        
        resource "hashicups_order" "test" {
          # (resource arguments)
        }
        
--- FAIL: TestAccOrderResource (0.64s)
```

### Steps to Reproduce

Launch `TestAccOrderResource` test

austinvalle:
## Related Issue

Closes #516
Supersedes #519 

## Description

#516 was a regression split across two PRs so it was a little tough to catch, as well as the behavior of `teststep.Config` is not obvious:
- https://github.com/hashicorp/terraform-plugin-testing/pull/442
	- Removed `teststep.Config` usage in favor of a string, however, the original `mergedConfig` is actually ignored when creating a new `teststep.Config`, which is confusing, but would essentially result in a `ConfigFile/Directory` test step being shaved down into just a string that has a `required_providers` block.
- https://github.com/hashicorp/terraform-plugin-testing/pull/494
	- Removed the check for reusing a prior test step config, because it was checking a value that was always populated (either with real config via `step.Config` or just a shaved down `required_providers` block via `step.ConfigFile/Directory`)

This PR reintroduces the usage of a `teststep.Config` over a raw string, adjusts some of the existing tests to cover this scenario, and adds a new one for the command import mode.

## Rollback Plan

No

### DIFF
--- a/.changes/unreleased/BUG FIXES-20250606-110444.yaml
+++ b/.changes/unreleased/BUG FIXES-20250606-110444.yaml
@@ -1,5 +1,5 @@
 kind: BUG FIXES
-body: 'importstate: Fixed bug where prior test config is not used for `ConfigFile` or `ConfigDirectory`'
+body: 'helper/resource: Fixed bug with import state mode where prior test config is not used for `ConfigFile` or `ConfigDirectory`'
 time: 2025-06-06T11:04:44.925152-04:00
 custom:
     Issue: "516"
--- a/.changes/unreleased/BUG FIXES-20250606-110444.yaml
+++ b/.changes/unreleased/BUG FIXES-20250606-110444.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'importstate: Fixed bug where prior test config is not used for `ConfigFile` or `ConfigDirectory`'
+time: 2025-06-06T11:04:44.925152-04:00
+custom:
+    Issue: "516"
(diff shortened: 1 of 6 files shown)


rainkwan (dismissed):
Looks sensible and logically sound to me.

PR #528, opened 2025-07-11
discovery: query mode

bbasata:
## Description

This pull request is not intended to be merged. It is a rugged, working proof-of-concept that identifies the wiring that would enable a `terraform query` acceptance test.

cc: @stephybun @rainkwan fyi

## Rollback Plan

- [x] If a change needs to be reverted, we will roll out an update to the code within 7 days.

## Changes to Security Controls

None

PR #531, opened 2025-07-21
Initial work on Query command for List Resource

rainkwan:
## Related Issue

Building off #528

## Description

Adding ListResource and ValidateListResourceConfig from List Resource to plugin testing.

## Rollback Plan

No


internal/plugintest/working_dir.go
```diff
@@ -151,6 +157,80 @@ func (wd *WorkingDir) SetConfig(ctx context.Context, cfg teststep.Config, vars c
...
+
+	fi, err := d.Readdir(-1)
+
+	if err != nil {
+		return err
+	}
+
+	for _, file := range fi {
+		if file.Mode().IsRegular() {
+			if filepath.Ext(file.Name()) == ".warioform" || filepath.Ext(file.Name()) == ".json" || filepath.Ext(file.Name()) == ".hcl" {
```
stephybun, line 183: Do you think it makes sense to be more specific with the file extension here as well, also
>.warioform

😬 

```suggestion
			if filepath.Ext(file.Name()) == ".warioform" || filepath.Ext(file.Name()) == ".json" || filepath.Ext(file.Name()) == ".tfquery.hcl" {
```

internal/teststep/file.go
```diff
@@ -71,6 +71,10 @@ func (c configurationFile) HasTerraformBlock(ctx context.Context) (bool, error)
 	return contains, nil
 }
 
+func (c configurationFile) WriteQuery(ctx context.Context, dest string) error {
+	panic("WriteQuery not supported for configurationDirectory")
```
stephybun, line 75:
```suggestion
	panic("WriteQuery not supported for configurationFile")
```

stephybun (approved):
Thanks @rainkwan, left a few minor comments but nothing that I think blocks us from merging this. LGTM 🚗

internal/testing/testsdk/providerserver/providerserver.go
```diff
@@ -12,12 +12,15 @@ import (
...
 
 	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/datasource"
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/list"
 	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/provider"
 	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/resource"
 )
 
 var _ tfprotov6.ProviderServer = ProviderServer{}
 
+// var _ tfprotov6.ProviderServerWithListResource = ProviderServer{}
```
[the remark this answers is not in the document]
austinvalle: Generally speaking, this interface should only be directly referenced by exported `terraform-plugin-go` packages or [testing packages like in `terraform-plugin-mux`](https://github.com/search?q=repo%3Ahashicorp%2Fterraform-plugin-mux%20ProviderServerWithListResource&type=code)

Since it's a temporary interface, we will eventually remove it once all downstream provider servers have a chance to implement them (for example, SDKv2 will need to implement the interface eventually, [even if it doesn't have an SDK for the RPCs](https://github.com/hashicorp/terraform-plugin-sdk/blob/8984e1e62a17bd607f069101a760e23a649ac2f8/helper/schema/grpc_provider.go#L1911-L1981)). Once that is removed, since there are no direct references to the interface itself, no downstream packages need to be re-released with the interface removed and provider's CI won't break 🙂

stephybun (approved):

PR #533, opened 2025-08-04
List Resource QueryCheck

rainkwan:
## Related Issue

Continues #531 

## Description
Adds QueryCheck to plugin testing to go along with Query

## Rollback Plan

No


querycheck/query_check.go
```diff
@@ -0,0 +1,33 @@
...
+// if the query file does not match what is expected.
+type QueryResultCheck interface {
+	// CheckQuery should perform the query check.
+	CheckQuery(context.Context, CheckQueryRequest, *CheckQueryResponse)
+}
+
+// CheckQueryRequest is a request for an invoke of the CheckQuery function.
+type CheckQueryRequest struct {
+	// Query represents a parsed query file, retrieved via the `terraform show -json` command.
+	Query *[]tfjson.ListResourceFoundData
```
austinvalle, line 22: nit: Also I don't think we need this to be a pointer to a slice here. Slices are already by reference under the hood

querycheck/query_check.go
```diff
@@ -0,0 +1,33 @@
...
+// QueryResultCheck defines an interface for implementing test logic that checks a query file and then returns an error
+// if the query file does not match what is expected.
+type QueryResultCheck interface {
+	// CheckQuery should perform the query check.
+	CheckQuery(context.Context, CheckQueryRequest, *CheckQueryResponse)
+}
+
+// CheckQueryRequest is a request for an invoke of the CheckQuery function.
+type CheckQueryRequest struct {
+	// Query represents a parsed query file, retrieved via the `terraform show -json` command.
```
austinvalle, line 21: nit: Are these the pre-parsed log messages or the results of parsing an actual file like the comment suggests

querycheck/expect_known_value.go
```diff
@@ -0,0 +1,74 @@
...
+	knownValue          knownvalue.Check
+}
+
+func (e expectKnownValue) CheckQuery(_ context.Context, req CheckQueryRequest, resp *CheckQueryResponse) {
+	for _, res := range *req.Query {
+		diags := make([]error, 0)
+
+		if e.listResourceAddress == strings.TrimPrefix(res.Address, "list.") && e.resourceName == res.DisplayName {
+			if res.ResourceObject == nil {
+				resp.Error = fmt.Errorf("%s - no resource object was returned, ensure `include_resource` has been set to `true` in the list resource config`", e.listResourceAddress)
```
austinvalle, line 30: nice catch! Feels like something a developer could easily run into 👍🏻

helper/resource/testing.go
```diff
@@ -640,6 +642,10 @@ type TestStep struct {
 	// Custom state checks can be created by implementing the [statecheck.StateCheck] interface, or by using a StateCheck implementation from the provided [statecheck] package.
 	ConfigStateChecks []statecheck.StateCheck
 
+	// ConfigQueryResultChecks allow assertions to be made against the query file during a Config test using a query check.
```
austinvalle, line 645: nit: Similar note to the other comment about package docs + naming. If we are inspecting a "query file" then i think the naming all makes sense (similar to inspecting a plan file or state file).

If we are inspecting logs from a query command, perhaps a different name would be better suited

helper/resource/testing.go
```diff
@@ -640,6 +642,10 @@ type TestStep struct {
...
+	// Custom query checks can be created by implementing the [querycheck.QueryResultCheck] interface, or by using a QueryResultCheck implementation from the provided [querycheck] package.
+	ConfigQueryResultChecks []querycheck.QueryResultCheck
```
austinvalle, line 647: nit: I think for naming, dropping `Config` might be best here. I think it's a little confusing to think of "query" as a part of the `Config` test step mode, since they serve completely different things.

I think it's okay to use the `Config` fields and not explicitly call the query mode part of the "config mode", if that makes sense. Similar to how you can use `Config` with the `ImportState` mode.

helper/resource/testing.go
```diff
@@ -855,6 +861,21 @@ type ConfigPlanChecks struct {
 	PostApplyPostRefresh []plancheck.PlanCheck
 }
 
+// ConfigQueryChecks defines the different points in a Config TestStep when query checks can be run.
+type ConfigQueryChecks struct {
```
austinvalle, line 865: nit: I don't think this is used anywhere? It seems to suggest that the query command would be run at different stages of the `Config` mode, but `Query` looks like it's own "test mode"

helper/resource/query/query_checks.go
```diff
@@ -0,0 +1,58 @@
...
+	"errors"
+	"fmt"
+
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/mitchellh/go-testing-interface"
+
+	"github.com/hashicorp/terraform-plugin-testing/querycheck"
+)
+
+func RunQueryChecks(ctx context.Context, t testing.T, query *[]tfjson.LogMsg, queryChecks []querycheck.QueryResultCheck) error {
```
austinvalle, line 17: nit: Similar note, slices don't need the pointer here and can just be `query []tfjson.LogMsg`

helper/resource/query/query_checks_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/querycheck"
+)
+
+var _ querycheck.QueryResultCheck = &queryCheckSpy{}
+
+type queryCheckSpy struct {
+	err    error
+	called bool
+}
```
austinvalle, line 17: nit: Doesn't look like this is used anywhere, perhaps it will eventually?

austinvalle (dismissed):
Looking good! Sounds like this is going to be a beta, so we can still come back and change any naming later if we so choose. All of the comments I left are nits that can be addressed/dismissed later

querycheck/expect_result_length_exact.go
```diff
@@ -0,0 +1,45 @@
...
+	if err := e.check.CheckValue(lengthCheck); err != nil {
+		resp.Error = fmt.Errorf("Query result of length %v - expected but got %v.", e.check, req.CompletedQuery.Total)
+		return
+	}
+}
+
+// ExpectLength returns a query check that asserts that the length of the query result is exactly the given value.
+//
+// This query check can only be used with managed resources that support query. Query is only supported in Terraform v1.14+
+func ExpectLength(resourceAddress string, length knownvalue.Check) QueryResultCheck {
```
gdavison, line 40: `ExpectLengthAtLeast` takes an `int` length value, but `ExpectLength` uses a `knownvalue.Check`. `int` seems like a better option

ansgarm (approved):

PR #553, opened 2025-09-19
querychecks: add tests

stephybun:
## Description

This PR 
- updates the schema/attributes that the examplecloud resource and list resource returns
- adds test files and test cases for the various query checks

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

### DIFF
--- a/helper/resource/query/query_test.go
+++ b/helper/resource/query/query_test.go
@@ -97,56 +97,6 @@ func TestQuery(t *testing.T) {
 					}),
 				},
 			},
-
-			// Commented out since these will fail now
-			//{
-			//	Query: true,
-			//	Config: `
-			//	provider "examplecloud" {}
-			//	list "examplecloud_containerette" "test" {
-			//		provider = examplecloud
-			//
-			//		config {
-			//			id = "westeurope/somevalue"
-			//		}
-			//	}
-			//	list "examplecloud_containerette" "test2" {
-			//		provider = examplecloud
-			//
-			//		config {
-			//			id = "foo"
-			//		}
-			//	}
-			//	`,
-			//	QueryResultChecks: []querycheck.QueryResultCheck{
-			//		querycheck.ExpectLength("examplecloud_containerette.test", 3),
-			//		querycheck.ExpectLength("examplecloud_containerette.test2", 3),
-			//	},
-			//},
-			//{
-			//	Query: true,
-			//	Config: `
-			//	provider "examplecloud" {}
-			//	list "examplecloud_containerette" "test" {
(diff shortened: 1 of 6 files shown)


querycheck/expect_known_value_test.go
```diff
@@ -0,0 +1,142 @@
...
+				`,
+				QueryResultChecks: []querycheck.QueryResultCheck{
+					querycheck.ExpectKnownValue(
+						"examplecloud_containerette.test",
+						"banana",
+						tfjsonpath.New("instances"),
+						knownvalue.NumberExact(big.NewFloat(5)),
+					),
+				},
+				ExpectError: regexp.MustCompile("examplecloud_containerette.test - the resource banana was not found"),
```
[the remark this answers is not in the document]
stephybun: There were actually some bugs in the test providerserver that we didn't catch earlier which were causing the issues. I pushed a commit to fix those and the test is passing now!

helper/resource/query/query_test.go
```diff
@@ -49,96 +53,100 @@ func TestQuery(t *testing.T) {
...
-						id = "foo"
-					}
-				}
-				`,
-				QueryResultChecks: []querycheck.QueryResultCheck{
-					querycheck.ExpectLengthAtLeast("examplecloud_containerette.test", 2),
-					querycheck.ExpectLengthAtLeast("examplecloud_containerette.test2", 1),
-				},
-			},
+			// Commented out since these will fail now
```
stephybun: Are these still irrelevant? If so let's remove them then.

querycheck/expect_known_value_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+					config {
+						resource_group_name = "foo"
+ 					}
+				}
+				`,
+				QueryResultChecks: []querycheck.QueryResultCheck{
+					querycheck.ExpectKnownValue(
+						"examplecloud_containerette.test",
+						"banana",
```
stephybun:
```suggestion
						"banane",
```

querycheck/expect_known_value_test.go
```diff
@@ -0,0 +1,143 @@
...
+						tfjsonpath.New("instances"),
+						knownvalue.NumberExact(big.NewFloat(4)),
+					),
+				},
+				ExpectError: regexp.MustCompile("examplecloud_containerette.test - the resource banana was not found"),
```
stephybun: This test should validate that the `ExpectKnownValue` check correctly flags values that are incorrect, but here we're checking for the presence of the resource.

PR #568, opened 2025-10-20
query: refactor query test step logic into `testStepNewQuery()`

SBGoods:
## Related Issue

Relates: #561 

## Description

Refactors query test step logic into `testStepNewQuery()` allowing query errors to be used with `ExpectError`.

## Rollback Plan

N/A

### DIFF
--- a/helper/resource/testing_new_query.go
+++ b/helper/resource/testing_new_query.go
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package resource
 
 import (


helper/resource/query/query_checks.go
```diff
@@ -28,6 +28,13 @@ func RunQueryChecks(ctx context.Context, t testing.T, query []tfjson.LogMsg, que
 
 	for _, msg := range query {
 		switch v := msg.(type) {
+		case tfjson.DiagnosticLogMessage:
+			if v.Diagnostic.Severity == tfjson.DiagnosticSeverityError {
+				// TODO reimplement missing .tf config error
+				result = append(result, fmt.Errorf("query error diagnostic - summary: %s, detail: %s", v.Diagnostic.Summary, v.Diagnostic.Detail))
+				continue
+			}
+
```
austinvalle: Not a blocking comment, but food for thought. One way this differs from other types of checks (plan / state checks) is that if any of the other commands fail, they immediately return an error (usually from the `wd` method) which is then checked by `ExpectError` without running any checks.

Is an error diagnostic returned by a provider considered the end of a `query` command? Meaning that  running query checks on any results after would be considered invalid? Another way to achieve the same result here would just be returning the error diagnostic from `func (wd *WorkingDir) Query(ctx context.Context) ([]tfjson.LogMsg, error)`

helper/resource/query/query_test.go
```diff
@@ -100,3 +104,88 @@ func TestQuery(t *testing.T) {
...
+
+				list "examplecloud_containerette" "test2" {
+					provider = examplecloud
+
+					config {
+						resource_group_name = "bar"
+					}
+				}
+				`,
+				ExpectError: regexp.MustCompile(`Diagnostic summary`),
```
austinvalle: > Something worth noting that might not be an immediate problem, but more of a potential future confusion for folks familiar with `ExpectError`

Since `Query` is using the machine readable output, the error message it is running the regex on with that mode is the formatted `%+v` of the diagnostics, which is different then other modes that use the formatted Terraform CLI output of the diagnostics. That can make the output hard to read / different to match than the other modes:
```bash
--- FAIL: TestQuery_ExpectError_ValidationError (0.49s)
    /Users/austin.valle/code/terraform-plugin-testing/helper/resource/query/query_test.go:111: Step 2/2 error running query, expected an error with pattern (Not the correct diagnostic summary), no match on: running terraform query command returned diagnostics: [{baseLogMessage:{Lvl:error Msg:Error: Diagnostic summary Time:2025-10-29 09:10:02.151561 -0400 EDT} Diagnostic:{Severity:error Summary:Diagnostic summary Detail:Diagnostic details Range:0x140004547c0 Snippet:0x14000030640}} {baseLogMessage:{Lvl:error Msg:Error: Diagnostic summary Time:2025-10-29 09:10:02.151735 -0400 EDT} Diagnostic:{Severity:error Summary:Diagnostic summary Detail:Diagnostic details Range:0x14000454900 Snippet:0x14000030690}}]
```

austinvalle (approved):
Just one non-blocking note, code LGTM 👍🏻

helper/resource/query/query_test.go
SBGoods: Yeah, that's a good point. Maybe we should introduce a string formatter for machine readable error messaging in a follow-up PR.

Issue #563, opened 2025-10-10
Proposal: Simplify `KnownValue` checks for query results


gdavison:
### terraform-plugin-testing version

```
github.com/hashicorp/terraform-plugin-testing v1.14.0-beta.1
```

### Use cases

When making `knownvalue` checks on a resource returned by a query, it will be more readable to perform one lookup for the resource and then apply one or more `knownvalue` checks.

This will especially be the case when looking up a resource by an identity with multiple attributes is supported.

### Proposal

Add a `querycheck` function, e.g. `ExpectResourceKnownValues`, which will take a list-resource address, a resource lookup mechanism (either by display name or resource identity), and one or more path and `knownvalue` check pairs.

For example:

```go
querycheck.ExpectResourceKnownValues("example_resource", querycheck.ResourceByIdentity(...), []querycheck.KnownValueCheck{
  querycheck.KnownValue(tfjsonpath.New("attr1"), knownvalue.NotNull()),
  querycheck.KnownValue(tfjsonpath.New("name"), knownvalue.StringExact("example")),
})
```

It will be an error if the lookup mechanism returns either zero or more than one result.

PR #573, opened 2025-11-11
Add `queryfilter` package and `ExpectResourceDisplayName` query check

### LINKED ISSUE #565: Proposal: Enhance `querycheck`s for resource display name

```
github.com/hashicorp/terraform-plugin-testing v1.14.0-beta.1
```

### Use cases

When testing query results, we want to be able to validate that a given resource has a specific display name.

In many instances, the display name is not knowable or not fully-knowable before run-time. For example, for many AWS EC2 resource types, the name is taken from the well-known tag `Name`, and therefore is not unique. This means that another value is appended to the display name for uniqueness. In other cases, resources do not have explicit names, so the display name is derived from one or more resource attributes which are randomly generated by AWS.

### Attempted solutions

The current display name check, `ContainsResourceWithName`, expects an exact match for the name and checks only for the existence of the name, not that it is associated with a specific resource.

### Proposal

Add a `querycheck` function, e.g. `ExpectResourceDisplayName`, that takes a list-resource address, a resource lookup mechanism, and allows passing a check for the display name. Display name checks could include exact match and regex match.

For example:

```go
querycheck.ExpectResourceDisplayName("example_resource", querycheck.ResourceByIdentity(...), querycheck.DisplayNameExact("expected name"))
```

```go
querycheck.ExpectResourceDisplayName("example_resource", querycheck.ResourceByIdentity(...), querycheck.DisplayNameRegexp(`expected name \([a-z0-9]+\)`))
```

Additionally, the display name check should allow customization and give access to all of the resource's attributes. This could allow, e.g. checking against a display name that contains `Computed` attributes.

For example:

```go
CustomDisplayNameFormat("{name} ({id})")
```

This example, `CustomDisplayNameFormat`, uses a similar format to the [`ExpectRegionalARNFormat`](https://github.com/hashicorp/terraform-provider-aws/blob/28b456121a4133599fe792acc4c235d8375b6939/internal/acctest/statecheck/expect_regional_arn_format.go#L51), [`ExpectGlobalARNFormat`](https://github.com/hashicorp/terraform-provider-aws/blob/28b456121a4133599fe792acc4c235d8375b6939/internal/acctest/statecheck/expect_global_arn_format.go#L49), and [`ExpectAttributeFormat`](https://github.com/hashicorp/terraform-provider-aws/blob/28b456121a4133599fe792acc4c235d8375b6939/internal/acctest/statecheck/expect_attribute_format.go#L52) functions used in the AWS provider

### References

The resource lookup mechanism, e.g. `querycheck.ResourceByIdentity(...)` was proposed in #563

SBGoods:
## Related Issue

Fixes #565 

## Description

Adds a new `queryfilter` package and `QueryFilter` interface for implementing declarative logic for filtering query results before the results are passed as input for a `QueryCheck`. This could be used to apply a check to a certain subset of results or as a lookup mechanism to check a specific result.

Filters are implemented at the query check level, so a single test step can have multiple independent query checks with their own filtering.

The current implementations of query filters are:
- `FilterByDisplayName`
- `FilterByResourceIdentity()`

This PR also introduces a new `ExpectResourceDisplayName()` query check that takes a filter and knownvalue.stringcheck for the display name as input and asserts that the display name matches the input after filtering.

## Rollback Plan

No

### DIFF
--- a/querycheck/expect_resource_display_name.go
+++ b/querycheck/expect_resource_display_name.go
@@ -20,7 +20,7 @@ var _ QueryResultCheckWithFilters = expectResourceDisplayName{}
 type expectResourceDisplayName struct {
 	listResourceAddress string
 	filter              queryfilter.QueryFilter
-	displayName         knownvalue.StringCheck
+	displayName         knownvalue.Check
 }
 
 func (e expectResourceDisplayName) QueryFilters(ctx context.Context) []queryfilter.QueryFilter {
@@ -51,7 +51,7 @@ func (e expectResourceDisplayName) CheckQuery(_ context.Context, req CheckQueryR
 		return
 	}
 	res := listRes[0]
-	if err := e.displayName.CheckString(res.DisplayName); err != nil {
+	if err := e.displayName.CheckValue(res.DisplayName); err != nil {
 		resp.Error = fmt.Errorf("error checking value for display name %s, err: %s", e.displayName.String(), err)
 		return
 	}
@@ -61,7 +61,7 @@ func (e expectResourceDisplayName) CheckQuery(_ context.Context, req CheckQueryR
 // ExpectResourceDisplayName returns a query check that asserts that a resource with a given display name exists within the returned results of the query.
 //
 // This query check can only be used with managed resources that support query. Query is only supported in Terraform v1.14+
-func ExpectResourceDisplayName(listResourceAddress string, filter queryfilter.QueryFilter, displayName knownvalue.StringCheck) QueryResultCheck {
+func ExpectResourceDisplayName(listResourceAddress string, filter queryfilter.QueryFilter, displayName knownvalue.Check) QueryResultCheck {
 	return expectResourceDisplayName{
 		listResourceAddress: listResourceAddress,
 		filter:              filter,
--- a/querycheck/queryfilter/filter_by_display_name.go
+++ b/querycheck/queryfilter/filter_by_display_name.go
@@ -10,11 +10,11 @@ import (
 )
 
 type filterByDisplayName struct {
-	displayNameCheck knownvalue.StringCheck
+	displayNameCheck knownvalue.Check
 }
 
 func (f filterByDisplayName) Filter(ctx context.Context, req FilterQueryRequest, resp *FilterQueryResponse) {
-	if err := f.displayNameCheck.CheckString(req.QueryItem.DisplayName); err == nil {
+	if err := f.displayNameCheck.CheckValue(req.QueryItem.DisplayName); err == nil {
 		resp.Include = true
 		return
 	}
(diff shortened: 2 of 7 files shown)


austinvalle (commented):
Summarizing a chat @SBGoods + I had offline, all of the code looks good to me here, we just want to make sure that the solution here meshes with how AWS was viewing the original problems in #565 + #563. More discussions to happen offline 👍🏻

stephybun (dismissed):
Thanks @SBGoods LGTM 💯 

Before we release I think we should also remove the `ContainsResourceWithName` check since the `ExpectResourceDisplayNameExact` check now supersedes it.

querycheck/queryfilter/filter_by_display_name.go
```diff
@@ -0,0 +1,29 @@
...
+package queryfilter
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+)
+
+type filterByDisplayName struct {
+	displayNameCheck knownvalue.StringCheck
```
austinvalle: nit: Similar comment to above about using `knownvalue.Check`, doesn't feel like too much of a stretch

querycheck/queryfilter/filter.go
```diff
@@ -0,0 +1,32 @@
...
+
+// QueryFilter defines an interface for implementing declarative filtering logic to apply to query results before
+// the results are passed to a query check request.
+type QueryFilter interface {
+	Filter(context.Context, FilterQueryRequest, *FilterQueryResponse)
+}
+
+// FilterQueryRequest is a request to a filter function.
+type FilterQueryRequest struct {
+	// Query represents the parsed log messages relating to found resources returned by the `terraform query -json` command.
```
austinvalle: nit: This just represents a single parsed log message
```suggestion
	// QueryItem represents a single parsed log message relating to a found resource returned by the `terraform query -json` command.
```

austinvalle (approved):

querycheck/expect_resource_display_name.go
```diff
@@ -10,19 +10,20 @@ import (
...
-var _ QueryResultCheckWithFilters = expectResourceDisplayNameExact{}
+var _ QueryResultCheck = expectResourceDisplayName{}
+var _ QueryResultCheckWithFilters = expectResourceDisplayName{}
 
-type expectResourceDisplayNameExact struct {
+type expectResourceDisplayName struct {
 	listResourceAddress string
 	filter              queryfilter.QueryFilter
-	displayName         string
+	displayName         knownvalue.StringCheck
```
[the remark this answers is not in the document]
austinvalle: Yeah my personal opinion is it's not too big of a deal. I think everyone involved with testing a display name understand it's a string, so it's only covering accidental usages which will immediately get an error when running the test anyways 🤷🏻

PR #583, opened 2025-12-02
Update `ExpectKnownValue` query check to use `queryfilter` and accept multiple `knownvalue` checks

### LINKED ISSUE #563: Proposal: Simplify `KnownValue` checks for query results

stephybun:
## Related Issue

Closes #563

## Description

This PR updates the existing `querycheck.ExpectKnownValue` check to be more consistent with the way a query check functions and is named after the introduction of a lookup mechanism introduced by #573

The following changes were made:
- The `ExpectKnownValue` check was updated to accept and use a `queryfilter` to find the relevant resource for assertions instead of the display name which may not always be unique or known beforehand
- The type `querycheck.KnownValuesCheck` was introduced to be able to specify multiple known value checks
- Check was renamed to `querycheck.ExpectResourceKnownValues` to better reflect what the check is capable of

## Rollback Plan

None.

### DIFF
--- a/querycheck/expect_resource_known_values.go
+++ b/querycheck/expect_resource_known_values.go
@@ -80,7 +80,6 @@ func (e expectResourceKnownValues) CheckQuery(_ context.Context, req CheckQueryR
 	}
 }
 
-// it's asserting that the resource object (identified by the filter) passes the given query checks".
 // ExpectResourceKnownValues returns a query check which asserts that a resource object identified by a query filter
 // passes the given query checks.
 //


querycheck/expect_resource_known_values.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+			resp.Error = fmt.Errorf("the following errors were found while checking values: %s", diagsStr)
+			return
+		}
+	}
+
+	resp.Error = fmt.Errorf("%s - the resource %s was not found", e.listResourceAddress, e.filter)
+}
+
+// ExpectResourceKnownValues returns a query check that asserts the specified attribute values are present for a given resource object
```
austinvalle: nit: Should this be reworded? I feel like it's more accurate to say something along the lines of "it's asserting that the resource object (identified by the filter) passes the given query checks". Since there is no specified attribute like `statecheck.KnownValue` or the previous query check implementation had.

querycheck/expect_resource_known_values.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+var _ QueryResultCheck = expectResourceKnownValues{}
+var _ QueryResultCheckWithFilters = expectResourceKnownValues{}
+
+type expectResourceKnownValues struct {
+	listResourceAddress string
+	filter              queryfilter.QueryFilter
+	knownValues         []KnownValueCheck
```
austinvalle: super nit 🦸🏻 - These are checks, not known values 🙂  , I think `knownValueChecks` would be a little easier to read further down in the code.

querycheck/known_value.go
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package querycheck
+
+import (
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+type KnownValueCheck struct {
```
austinvalle: Since this is an external package, a little documentation blurb about what this struct is / how it relates to resource objects found via query filters, might be useful 👍🏻

ansgarm (dismissed):
image

ansgarm (approved):

PR #604, opened 2026-02-09
query: fix expect length query check by collecting all query summaries and matching on resource address

stephybun:
## Related Issue

Related to #597

## Description

This is the first PR that addresses the problems outlined in #597 it does the following:
- Introduces the `QuerySummaries` property to collect all available `ListCompleteMessage`, if there are multiple list blocks in the config then there are multiple `ListCompleteMessage`
- Deprecated the `QuerySummary` property
- Updated the behaviour of the `ExpectLength` query check to first match on the resource address before comparing the count
- Duplicated the mock resource and list resource to add the `examplecloud_bananette` resource to cover additional test cases for this fix

Up Next:
- A separate PR adding support for regex based matching to calculate counts from dynamically created list blocks
- Updating the user facing documentation on query checks with the changes from this PR and the new check described in the point above

## Rollback Plan

None

### DIFF
--- a/querycheck/expect_result_length_exact.go
+++ b/querycheck/expect_result_length_exact.go
@@ -18,7 +18,7 @@ type expectLength struct {
 
 // CheckQuery implements the query check logic.
 func (e expectLength) CheckQuery(_ context.Context, req CheckQueryRequest, resp *CheckQueryResponse) {
-	if req.QuerySummary == nil && req.QuerySummaries == nil {
+	if req.QuerySummary == nil && len(req.QuerySummaries) == 0 {
 		resp.Error = fmt.Errorf("no query summary information available")
 		return
 	}


querycheck/expect_result_length_exact.go
```diff
@@ -17,15 +18,31 @@ type expectLength struct {
 
 // CheckQuery implements the query check logic.
 func (e expectLength) CheckQuery(_ context.Context, req CheckQueryRequest, resp *CheckQueryResponse) {
-	if req.QuerySummary == nil {
+	if req.QuerySummary == nil && req.QuerySummaries == nil {
```
austinvalle: Should the 2nd part of the conditional be `len(req.QuerySummaries) == 0`?

austinvalle (approved):

PR #633, opened 2026-03-17
Single Workspace Skip - DefaultWorkspaceOnly

rainkwan:
## Related Issue

Fixes # 

## Description

This PR adds a DefaultWorkspaceOnly boolean to TestStep that makes it so tests are limited to the provider's default workspace, a special case that applies to testing StateStores

## Rollback Plan

No


helper/resource/testing.go
```diff
@@ -861,6 +861,12 @@ type TestStep struct {
 	//   - Supports workspaces (creating and deleting)
 	StateStore bool
 
+	// DefaultWorkspaceOnly limits StateStore mode tests to the provider's default workspace.
+	// This should only be used in rare cases where exercising workspace switching is unnecessary
+	// for validating the state store implementation. Prefer tests that exercise workspace handling
+	// unless there is a specific reason to restrict testing to the default workspace.
+	DefaultWorkspaceOnly bool
```
austinvalle: Since this is only related to `StateStore`, perhaps additional validation might be useful for `TestStep` to enforce that? 👀

.changes/unreleased/ENHANCEMENTS-20260319-131506.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 
+helper/resource: Added `DefaultWorkspaceOnly` to enable testing state store implementations that don't support multiple workspaces. This is achieved by only running Terraform commands with the `default` workspace.
```
austinvalle: It's weird changie didn't raise an error here, as this looks like an invalid changelog 🤔 , it looks like the body would be empty because of the newline before `helper/resource:`

Image

austinvalle (dismissed):
Looks like there is a small format change to the changelog needed, but everything else LGTM 🚀

.changes/unreleased/ENHANCEMENTS-20260319-131506.yaml
rainkwan: Yeah for some reason if I do not put a new line it gets angry because it thinks that section needs a new header so maybe if I just format it differently?

austinvalle: Yeah probably because of the `:` character, wrapping the entire thing in single quotes will treat it as a string literal in YAML, most of our changelogs have something similar: https://github.com/hashicorp/terraform-plugin-testing/blob/c5d6dddcb365253c93d49891cd74e4136f61c53b/.changes/unreleased/BUG%20FIXES-20260211-103028.yaml#L2

rainkwan: I tried that once before but then changie fails for some reason?

austinvalle: Very interesting 🤔, it's likely an encoding problem 👀

.changes/unreleased/ENHANCEMENTS-20260319-131506.yaml
```diff
@@ -0,0 +1,5 @@
+kind: ENHANCEMENTS
+body: 'helper/resource: Added `DefaultWorkspaceOnly` to enable testing state store implementations that don't support multiple workspaces. This is achieved by only running Terraform commands with the `default` workspace.'
```
austinvalle:
```suggestion
body: 'helper/resource: Added `DefaultWorkspaceOnly` to enable testing state store implementations that don''t support multiple workspaces. This is achieved by only running Terraform commands with the `default` workspace.'
```

austinvalle: @rainkwan I think this is it 😆, single quote escaping is just `''` -> `'`, YAML is fun 😵

austinvalle (approved):

PR #638, opened 2026-03-25
Add `GenerateConfig` testing mode to `ImportState` testing

SBGoods:
## Description

Adds a `GenerateConfig` testing mode to import mode testing. This mode runs an import plan with `-generate-config-out` flag and throws an error if there is a diff produced in the plan when compared to the state in the previous step.

## Rollback Plan

N/A

### DIFF
--- a/helper/resource/importstate/import_block_with_id_test.go
+++ b/helper/resource/importstate/import_block_with_id_test.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-testing/internal/testing/testsdk/provider"
 	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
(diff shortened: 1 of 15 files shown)


helper/resource/testing_new_import_state.go
```diff
@@ -158,11 +170,13 @@ func testStepNewImportState(ctx context.Context, t testing.T, helper *plugintest
 				t.Fatalf("copying state: %s", err)
 			}
 
-			err = runProviderCommand(ctx, t, workingDir, providers, func() error {
-				return workingDir.RemoveResource(ctx, resourceName)
-			})
-			if err != nil {
-				t.Fatalf("removing resource %s from copied state: %s", resourceName, err)
+			if !step.GenerateConfig {
```
ansgarm, line 173: can we add a command why we don't delete the resource in this case? Is it deleted later through some cleanup or something instead?

helper/resource/importstate/import_block_with_id_test.go
```diff
@@ -487,3 +488,73 @@ import {
...
+
+	r.UnitTest(t, r.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.SkipBelow(tfversion.Version1_14_0), // Provider generated config requires Terraform 1.14.0 or later
+		},
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"examplecloud": providerserver.NewProviderServer(testprovider.Provider{
+				Capabilities: &provider.ServerCapabilities{
+					// GenerateResourceConfig RPC is unimplemented, returning a null generated
+					// config to Terraform Core.
```
ansgarm, line 535: good comment!

ansgarm (approved):
Looks good to me! Only one small question but that can be followed-up / your choice whether to address :)

